PR #28, opened 2016-11-21
Options updates

### LINKED ISSUE #26: Some issues with `gk-deploy`
Long arguments seem to have issues:
```
$ ./gk-deploy --deploy-gluster --verbose                                                                                               
Unknown option 'deploy-gluster'.

$ ./gk-deploy -g --verbose                                                                                                          1  
Unknown option 'erbose'.
```
Short versions work, but there seem to be some other issues:
```
./gk-deploy -g -v                                                                                                                    
Starting Kubernetes deployment.
serviceaccount "heketi-service-account" created
Found secret 'heketi-service-account-token-3juax' in namespace 'default' for heketi-service-account.
  File "<stdin>", line 10
    print node['node']['hostnames']['manage'][0]
             ^
SyntaxError: Missing parentheses in call to 'print'
Deploying GlusterFS pods on .
The Deployment "glusterfs-" is invalid: metadata.name: Invalid value: "glusterfs-": must match the regex [a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)* (e.g. 'example.com')
Waiting for GlusterFS pods to start ... Checking status of pods matching 'glusterfs=pod':

Checking status of pods matching 'glusterfs=pod':
[...]
```

### LINKED ISSUE #27: Use namespace for heketi/gluster resources
If all resources of Heketi and GlusterFS reside in one namespace, it might be easier to filter for them on metrics or logging.

Also to stop and remove all resources gracefully you then can just call:
```
kubectl delete namespace glusterfs
```

jarrpa:
Fix command-line option parsing and add namespace option.

Closes #26 
Closes #27

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -21,11 +21,13 @@ VERBOSE=0
 CLI=''
 GLUSTER=0
 TEMPLATES="templates"
+NAMESPACE="default"
 WAIT=300
 ABORT=0
 
 usage() {
-  echo -e "USAGE: ${PROG} [-ghv] [-t <TEMPLATES>] [-w <SECONDS>] [-l <LOG_FILE>] [<TOPOLOGY>]\n"
+  echo -e "USAGE: ${PROG} [-ghv] [-t <TEMPLATES>] [-n NAMESPACE] [-w <SECONDS>]
+       [-l <LOG_FILE>] [<TOPOLOGY>]\n"
 }
 
 help_exit() {
@@ -49,6 +51,10 @@ Options:
   -t TEMPLATES, --templates_dir TEMPLATES
               Location of directory containing the heketi templates for
               OpenShift. Default is '${TEMPLATES}'.
+
+  -n NAMESPACE, --namespace NAMESPACE
+              The namespace to use for creating resources.
+              Defaults is '${NAMESPACE}'.
 
   -w SECONDS, --wait SECONDS
               Wait SECONDS seconds for pods to become ready. Default is '${WAIT}'.
@@ -85,12 +91,12 @@ debug() {
 }
 
 abort() {
-  kubectl delete all,service,jobs,deployment,secret --selector="deploy-heketi" 
-  kubectl delete all,service,jobs,deployment,secret --selector="heketi"
-  kubectl delete svc/heketi-storage-endpoints
-  kubectl delete sa heketi-service-account
+  ${CLI} delete all,service,jobs,deployment,secret --selector="deploy-heketi"
+  ${CLI} delete all,service,jobs,deployment,secret --selector="heketi"
+  ${CLI} delete svc/heketi-storage-endpoints
+  ${CLI} delete sa heketi-service-account
   if [[ ${GLUSTER} -eq 1 ]]; then
-    kubectl delete all,service,jobs,deployment,secret --selector="glusterfs"
+    ${CLI} delete all,service,jobs,deployment,secret --selector="glusterfs"
   fi
   exit 1
 }
@@ -124,7 +130,7 @@ check_pods() {
     fi
     sleep 2
     ((s+=2))
-    pods=$(kubectl get pod --no-headers --show-all --selector="${1}")
+    pods=$(${CLI} get pod --no-headers --show-all --selector="${1}")
     debug "Checking status of pods matching '${1}':"
     debug "$pods"
     all_ready=true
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -40,20 +43,26 @@ Arguments:
...
+              or removal.
 
   -t TEMPLATES, --templates_dir TEMPLATES
               Location of directory containing the heketi templates for
               OpenShift. Default is '${TEMPLATES}'.
 
+  -n NAMESPACE, --namespace NAMESPACE
+              The namespace to use for creating resources.
+              Defaults is '${NAMESPACE}'.
+
```
humblec: We normally place everything in one namespace, However we can still mention different namespace for a particular resource . Isnt it ? have we tested it ?

deploy/gk-deploy
```diff
@@ -40,20 +43,26 @@ Arguments:
...
   -w SECONDS, --wait SECONDS
               Wait SECONDS seconds for pods to become ready. Default is '${WAIT}'.
 
   -l LOG_FILE, --log LOG_FILE
               Save all output to the specified file.
 
-  --abort     Abort a failed deployment. WARNING: Deletes all related resources.
+  --abort     Abort a deployment. WARNING: Deletes all related resources.
```
humblec: Instead of a complete removal, it would good to give an option or checkpoint based removal. For ex: if the gluster pods are up and running, we may not delete those and recreate it.. Thus we could have different checkpoints and receive inputs from the user and proceed. This could help us to rollback to a particular state and proceed from there.

deploy/gk-deploy
```diff
@@ -40,20 +43,26 @@ Arguments:
...
+              brick devices. If the --abort flag is also specified, this flag
+              indicates that all GlusterFS pods and deployments should be
+              deleted as well. Default is to not handle GlusterFS deployment
+              or removal.
 
   -t TEMPLATES, --templates_dir TEMPLATES
               Location of directory containing the heketi templates for
               OpenShift. Default is '${TEMPLATES}'.
 
+  -n NAMESPACE, --namespace NAMESPACE
```
humblec: We could also create a specified namespace if required.. Its very easy or a single command if that namespace doesnt exist.

deploy/gk-deploy
```diff
@@ -308,29 +330,29 @@ fi
...
 
-if [ "${CLI}" == "oc" ]; then
-  oc process heketi -v \
+if [ "${CLI}" == *oc* ]; then
+  ${CLI} process heketi -v \
          HEKETI_KUBE_NAMESPACE=heketi \
          HEKETI_KUBE_INSECURE=y \
          HEEKTI_KUBE_SECRETNAME="${heketi_secret}" \
-         HEKETI_KUBE_APIHOST="${kubeapi}" | oc create -f -
+         HEKETI_KUBE_APIHOST="${kubeapi}" | ${CLI} create -f -
```
humblec: These (HEKETI_*) should be declared as constants.

obnoxxx:
@humblec are the above comments requests for change? If so I think you should now have rights to raise this review flag.

deploy/gk-deploy
jarrpa: Constants are mainly just conventions in bash scripts. There's no way to hard code a constant value for this, as it would be different for each environment. if you mean you want the variables to be in all uppercase, that's a cosmetic issue I don't feel a need to address, as I reserve that for values that are non-generated or have static defaults (see the top of file).

deploy/gk-deploy
jarrpa: Does it make sense to do that? I don't want to make an assumption that a namespace should be created if it doesn't exist. It might make more sense to do a check to see if it exists and then fail if it doesn't.

jarrpa:
Please ignore the travis-ci notice, I was experimenting with using it, but I don't think it'll work for this project.

deploy/gk-deploy
humblec: @jarrpa

deploy/gk-deploy
humblec: @jarrpa my point is that, if admin request to create the entire project in a namespace,  do it. may be @obnoxxx can share his thoughts.

deploy/gk-deploy
humblec: @jarrpa may be I used wrong word, but my point is that, if you declare those as kind of global variables, it will be easy to maintain. for ex: HEKETI_KUBE_APIHOST may change to something tomo, we just need a change in one place.

humblec:
@obnoxxx yes, I requested for a change. Yep, now I have the permission. Thanks!

obnoxxx:
@humblec there is no request for change marked on this PR.

deploy/gk-deploy
```diff
@@ -214,24 +234,26 @@ if [ "x${CLI}" == "x" ]; then
...
-if [ "${CLI}" == "oc" ]; then
+CLI="${CLI} -n ${namespace}"
+
+if [ "${CLI}" == *oc* ]; then
   output "Starting OpenShift deployment."
-  oc new-project heketi
+  ${CLI} new-project heketi
   oadm policy add-scc-to-user privileged -z default
-  oc create -f /usr/share/heketi/templates
+  ${CLI} create -f /usr/share/heketi/templates
```
obnoxxx: @humblec maybe "could be" or "should be", but that is actually a completely independent request, right? We should not require to fix/improve everything we can think of in this patch. :-)

deploy/gk-deploy
obnoxxx: @humblec I'd also say:  A very useful proposal for an RFE, but not within the scope of this PR. Should we raise a new issue for this?

deploy/gk-deploy
obnoxxx: ==> new pr / issue?

deploy/gk-deploy
obnoxxx: Also this seems to be a general comment on the script, not so much a comment on the changes of this patch, right?

deploy/gk-deploy
jarrpa: I can put in a check to see if a namespace exists, at least. The Administrator should be left to create any missing namespace themselves. Making too many assumptions about and affecting the environment is outside the scope of this script.

deploy/gk-deploy
jarrpa: Please open a new Issue, I'll take care of it when I actually do the openshift integration properly. :)

deploy/gk-deploy
obnoxxx: @jarrpa agreed. I think the script does not need to expose every single bit of flexibility that the manual process offers. If we need it in the future, we can still add the flexibility....

deploy/gk-deploy
obnoxxx: @jarrpa, i see @humblec's point, but I agree that it would be a good first approach to have a check whether the namespace exists and fail if it does not.

deploy/gk-deploy
obnoxxx: @humblec, agree to @jarrpa : this is not an issue.

obnoxxx:
@jarrpa , summarizing, most of @humblec 's comments should be separate RFEs.
On thing about the creation of the namsepace is pending. If we have a check, then we should merge the patch and move on, imho.

jarrpa:
Updated the commits to add a namespace existence check.

obnoxxx:
@jarrpa 👍  Will merge unless I hear objections from humble tomorrow.

PR #30, opened 2016-11-27
Minor fixes to scripting

jarrpa:
Noticed a few things to clean up, including waiting for service account secrets to populate.

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -248,7 +248,11 @@ else
 fi
 
 ${CLI} create -f ${TEMPLATES}/heketi-service-account.yaml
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -162,13 +162,9 @@ while [[ $# -ge 1 ]]; do
     keypos=1
     while [[ $keypos -lt $keylen ]]; do
       case ${key:${keypos}} in
-        o*|-openshift)
-        CLI="oc"
-        if [ "$key" == "--openshift" ]; then keypos=$keylen; fi
-        ;;
```
obnoxxx: This removal hunk is (arguably) not a syntax cleanup. :-) What's the reasoning? Should it be a separate commit if at all?

jarrpa: It's a cleanup. I forgot to remove it earlier when I closed off openshift support and will currently break things if used. Hence, fix.

PR #34, opened 2016-11-30
OpenShift support, minor fixes

jarrpa:
The first two commits are small fixes I found along the way to writing the third commit, which implements OpenShift support for the deployment script. Said commit also moves around some files to make it clear which templates are for which container platform.


jarrpa:
For some reason the commits are display out of order in the GitHub interface. Their proper order (and I believe this is maintained in the git history) is:

```
be4464e Add OpenShift support.
7c19df8 Improve search for heketi-service-account secret.
12df060 Wait for heketi service URL to be active.
```

deploy/gk-deploy
```diff
@@ -20,13 +20,15 @@ LOG_FILE=''
 VERBOSE=0
 CLI=''
 GLUSTER=0
-TEMPLATES="templates"
+KUBE_TEMPLATES_DEFAULT="./kube-templates"
+OCP_TEMPLATES_DEFAULT="./ocp-templates"
+TEMPLATES="openshift"
```
obnoxxx: why default to openshift instead of kubernetes?

really, "openshift" does not seem correct ? should it be ocp-templates or kube-templates ?

jarrpa: That's a dev shortcut I forgot to remove. 'openshift' was the name of the directory I was using for templates, so I hard-coded it cause I was lazy. :)

jarrpa:
Oh... it seems when I push override a new commit over the old one, any comments that were made inline get deleted. Sorry. :(

@obnoxxx asked why I had a line `TEMPLATE="openshift"`, I explained that that was a development hard-coded shortcut and fixed it.

obnoxxx:
@jarrpa my comment is still there... :-)

PR #40, opened 2016-12-02
Update Install instructions

wattsteve:
- fix to say we are deploying gluster-kuberenetes rather than heketi
- fix to add sudo in the gk-deploy command example

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ For ease of use in the the vagrant setup, we recommend you run the following:
 $ export KUBECONFIG="/etc/kubernetes/admin.conf"
 ```
 
-Next, to deploy heketi, run the following:
+Next, to deploy gluster-kubernetes, run the following:
 
 ```bash
-$ ./gk-deploy -g
+$ sudo ./gk-deploy -g
 ```
 
 If you already have GlusterFS deployed in your cluster, you do not need the


jarrpa:
`gluster-kubernetes` is the name of the project to bring GlusterFS into Kubernetes, not the actual service being deployed. We're deploying heketi, and optionally GlusterFS (via the `-g` option). I do see the need to clarify the documentation for running with admin rights, however, so I'll do that.

humblec:
@jarrpa  as @wattsteve mentioned, this deployment tool is about deploying `gluster on kubernetes` and `heketi` comes as one  service in the project. So, I do think the PR is valid.

jarrpa:
@humblec `gluster on kubernetes` isn't accurate, and in fact the deployment of GlusterFS is a secondary (and OPTIONAL) step of the deployment script, as you can have your own GlusterFS pods or services running already. The *project* is about Gluster on Kubernetes, the *script* is about deploying heketi (and optionally GlusterFS).

obnoxxx:
@jarrpa, I tend to agree with @wattsteve in that the project consists of serveral services, and deployment of the gluster-kubernetes system or project means deploying heketi and (optionally) gluster. So yeah, it may be more natural to say .... "to deploy gluster-kubernetes"...

obnoxxx:
this needs to be rebased

jarrpa:
@obnoxxx Yes the project is about several services thus, again, there is no such 'gluster-kubernetes' thing to deploy. The main thing being deployed here is heketi, with a basic deployment of GlusterFS in pods as an optional step. It does not feel more natural, to me it sounds disingenuous to not explicitly call out heketi in that step.

obnoxxx:
@jarrpa ok... So you are saying that one can only deploy a single service, not a combination of multiple services? This is really about the correct use of the word "to deploy" to me:

I thought I could deploy e.g. a "gluster cluster" which amounts in deploying a number of storage nodes. In the same line, i would be possible to talk about deploying a combination of a heketi service and one or more gluster glusters, aka an instance of gluster-kuberenetes?

obnoxxx:
@jarrpa of course I am not saying we should *not* mention heketi, we certainly should!

jarrpa:
@obnoxxx I don't think we can really talk about an 'instance, service, cluster, or even deployment of "gluster-kubernetes". It's more a state of infrastructure, just because of the variety of configurations we can encounter. The only concrete infrastructure piece we mandate for this project is a Kubernetes cluster to run the heketi resources (e.g. pods, service). GlusterFS can be its own standalone cluster, or a preconfigured Kubernetes clustered service, or deployed directly via gk-deploy.

jarrpa:
After a meeting with @wattsteve, I've come around somewhat on the issue this PR is addressing. The documentation needs to be more broadly updated to reflect the current core mission of this repo, which is to provide a hyper-converged solution of GlusterFS on Kubernetes. That said, the changes in this PR are still not quite correct, so I'm closing this and will submit a broader documentation update shortly.

PR #41, opened 2016-12-05
Small ansible updates

jarrpa:
Some small ansible updates I've put in as a workaround for being on a slow Internet connection. The documentation updates are inspired by and supersede those in #40.

### DIFF
--- a/vagrant/roles/common/tasks/main.yml
+++ b/vagrant/roles/common/tasks/main.yml
@@ -8,6 +8,7 @@
 
 - name: disable of selinux - now
   command: setenforce 0
+  ignore_errors: yes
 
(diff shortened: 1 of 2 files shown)


obnoxxx:
@jarrpa it conflicts... needs rebase?

obnoxxx:
As mentioned in #40, I think the point of the other PR is valid. Let's merge the other patch first.

That's why I prefer having one PR stick to one topic. Makes life easier. ;-)

jarrpa:
In the interest of slimming down the PR, I've pushed an update to remove the second of the two commits, so we only have the ansible-related changes to consider.

Issue #50, opened 2016-12-07
Use current namespace instead of 'default'


jarrpa:
Currently, the default behavior of `gk-deploy` is to deploy into the namespace 'default'. We should instead deploy to whatever the user's current namespace is (e.g. not require a `-n` option).

PR #57, opened 2016-12-08
Namespace:user add to given namespace

MohamedAshiqrh:


jarrpa:
This does not close #50, as it does not actually use the current namespace by default. Please remove the "closes: #50" line form the PR.

jarrpa:
Updated the original PR message so it doesn't close #50. Merging.

PR #68, opened 2016-12-10
Major documentation update

jarrpa:
Major updates to the README and setup-guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,16 @@ deployment scheme.
 * **GlusterFS** (https://www.gluster.org/), the scale-out storage system, running as pods inside the kubernetes cluster.
 * **heketi** (https://github.com/heketi/heketi), Gluster's volume management service interface, running in a pod inside kubernetes.
 
+### Demo
+
+**>>> [Video demo of the technology!](https://drive.google.com/file/d/0B667S2caJiy7QVpzVVFNQVdyaVE/view?usp=sharing) <<<**
+
+### Documentation
+
+* [Quickstart](#quickstart)
+* [Setup Guide](./docs/setup-guide.md)
+* [Hello World with GlusterFS Dynamic Provisioning](./docs/examples/hello_world/README.md)
+
 ### Quickstart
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,17 +1,16 @@
 # gluster-kubernetes
 
-## Dynamic, persistent storage for Kubernetes with GlusterFS
+## GlusterFS + heketi on Kubernetes
```
obnoxxx: I disagree with this change of description.
The original description conveys what the project wants to achieve, the new description tells which components are used.

README.md
```diff
@@ -1,17 +1,16 @@
...
 
-This is the home of the **gluster-kubernetes** project,
-a new project which brings dynamically provisioned persistent
-storage volumes to kubernetes, using a hyperconverged Gluster
-deployment scheme.
+**gluster-kubernetes** is a project to provide Kubernetes administrators a
+mechanism for deploying GlusterFS and heketi onto their clusters.
```
obnoxxx: Same point of criticism as above. Exchanging the higher level mission with the lower level technical way to achieve it.

README.md
jarrpa: Apparently we're not the home for dynamic, persistent storage since we don't host the GlusterFS volume plugin for Kubernetes nor heketi itself, and those two things are what provide the dynamic provisioning and persistent volume support for GlusterFS. I can tack on the phrase "Hyper-converged" at the beginning, I suppose?

obnoxxx: Hm, well, we host none of the components, but this is the project that glues them all together to achieve that broader mission of providing that dynamic, persistent storage for containers, is my understanding. We achieve this by providing the glue code to comfortably deploy the needed resources. I think the proposed headline is not a lot less deceiving than the current one since it does not mention that we only provide means of deploying those, but could make one think that we are actually providing gluster and heketi....

obnoxxx: What I am missing in the proposed headline is the *purpuse* of this project. But we should not ly either way. I'll merge and we can then still discuss about a better headline........... :-)

PR #79, opened 2016-12-15
TOPOLOGY file check should not be happening, If ABORT is requested

MohamedAshiqrh:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/79)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -239,7 +239,7 @@ while [[ $# -ge 1 ]]; do
   shift
 done
 
-if [[ ! -f ${TOPOLOGY} ]]; then
+if [[ ${ABORT} -ne 1 ]] && [[ ! -f ${TOPOLOGY} ]]; then
   echo "Topology File not found!"
   exit 1
 fi


deploy/gk-deploy
```diff
@@ -239,7 +239,7 @@ while [[ $# -ge 1 ]]; do
   shift
 done
 
-if [[ ! -f ${TOPOLOGY} ]]; then
+if [[ ${ABORT} -ne 1 ]] && [[ ! -f ${TOPOLOGY} ]]; then
```
jarrpa: No need for an extra check, just move the entire topology check block to https://github.com/gluster/gluster-kubernetes/blob/master/deploy/gk-deploy#L303 just after the main abort check.

MohamedAshiqrh: Will it be useful. If I move to down, We will be asking for Namespace, cli and add template and then we will be quiting for topology file not found. I don't see a point why go through all these and then quit for topology file.

jarrpa: Okay, fair. However, we still have to worry about the solution to #81 because as it stands it introduces a need for the topology file in `abort()` thus these two PRs conflict with each other.

MohamedAshiqrh: @jarrpa I agree, so Will grep the labels in #81 .  Abort is a risky operation and user should know it is gonna remove all gluster when used.

jarrpa:
Seeing as we want to go forward with #81, shall we close this PR?

MohamedAshiqrh:
Only problem closing is gk-deploy --abort will still ask for a topology file. There is only one case where topology file is not required, Thats when you do # gk-deploy --abort this will just clean heketi.

jarrpa:
So we need the topology file in all cases EXCEPT when we are aborting without `-g`. I don't think this corner case with worth worrying about: `--abort` should only ever be used AFTER a deploy attempt started and failed (or was interrupted), in which case you would have had an extant topology file already.

MohamedAshiqrh:
Agreed, So closing this PR. TOPOLOGY file is always mandatory for deployment.

PR #81, opened 2016-12-15
Storagenode label remove on ABORT

MohamedAshiqrh:
Closes: #55 

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/81)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -26,6 +26,7 @@ TEMPLATES=""
 NAMESPACE="default"
 WAIT=300
 ABORT=0
+NODES=""
 
 usage() {
   echo -e "USAGE: ${PROG} [-ghv] [-c CLI] [-t <TEMPLATES>] [-n NAMESPACE] [-w <SECONDS>]
@@ -108,6 +109,10 @@ abort() {
     ${CLI} delete template heketi
   fi
   if [[ ${GLUSTER} -eq 1 ]]; then
+    while read -r node; do
+      debug "Removing label from '${node}' as a GlusterFS node."
+      ${CLI} label nodes ${node} storagenode-
+    done <<< "$(echo -e "${NODES}")"
     ${CLI} delete all,service,jobs,ds,secret --selector="glusterfs"
     if [[ "${CLI}" == *oc* ]]; then
       ${CLI} delete template glusterfs
@@ -242,6 +247,20 @@ done
 if [[ ! -f ${TOPOLOGY} ]]; then
   echo "Topology File not found!"
   exit 1
+else
+  NODES=$(python - <<END
+import sys
+import json
+import argparse
+
+file = open('${TOPOLOGY}', 'r')
+topo = json.load(file)
+
+for cluster in topo['clusters']:
+  for node in cluster['nodes']:
+    print(str(node['node']['hostnames']['manage'][0]))
+END
+)
 fi
 
 if [[ "x${CLI}" == "x" ]]; then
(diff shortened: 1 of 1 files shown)


MohamedAshiqrh:
sorry removed reviewable while editing the comment.

jarrpa:
This won't work because of #79. Given how we just arbitrarily search for the label anyway, you could do a `${CLI} get nodes` and just cycle through all nodes and remove the label (suppressing error messages if it doesn't exist on that node).

MohamedAshiqrh:
@jarrpa I thought of it too. But the problem what I thought was, what if they have nodes with storagenode=ceph (or any other name).

jarrpa:
Can't you do a search for the whole label's key=value pair?

MohamedAshiqrh:
@jarrpa I might think too much. There is a possibility that someone wants to deploy more than one cns cluster in the their HA openshift with different namespace but node labels will not change right?

```
[root@master ~]# oc project
Using project "aplo" on server "https://192.168.50.5:8443".
[root@master ~]# oc get nodes --show-labels
NAME           STATUS                     AGE       LABELS
192.168.50.2   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.2,storagenode=glusterfs
192.168.50.3   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.3,storagenode=glusterfs
192.168.50.4   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.4,storagenode=glusterfs
192.168.50.5   Ready,SchedulingDisabled   12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.5
[root@master ~]# oc project default
Now using project "default" on server "https://192.168.50.5:8443".
[root@master ~]# oc get nodes --show-labels
NAME           STATUS                     AGE       LABELS
192.168.50.2   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.2,storagenode=glusterfs
192.168.50.3   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.3,storagenode=glusterfs
192.168.50.4   Ready                      12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.4,storagenode=glusterfs
192.168.50.5   Ready,SchedulingDisabled   12d       beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/hostname=192.168.50.5
[root@master ~]# 
```

MohamedAshiqrh:
@jarrpa As per our discussion with respect to #79 Updating this PR with nessecary changes.

jarrpa:
There is no cns upstream. ;)

If we're considering the use case where we have multiple clusters managed by multiple instances of heketi in different namespaces, what should we do in the case where two clusters from different namespaces are on the same node? You can't have two GlusterFS pods on the same node, best I understand, so we should have a way to check for this and fail out. We should raise this as a separate issue.

In any case, I think you're right that we should only specifically unlabel the nodes in our topology. However, I don't like the idea of just copying the python scriptlet into two places. I'd like to see the script [just after the topology check](https://github.com/gluster/gluster-kubernetes/blob/master/deploy/gk-deploy#L246), with the output being set to a global variable called `NODES`. We can then use this global var both when we label and unlabel the nodes, and possibly other places as they call for it.

humblec:
LGTM

jarrpa:
Same! Excellent work, Ashiq. :)

PR #97, opened 2016-12-19
Fix Logging to file

ramkrsna:
- This patch fixes logging to a given file with the --log options along with appending to the log file.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/97)


ramkrsna:
As decided I'll the log-file option intact, and change the documentation, to use the --log-file instead of --log option

ramkrsna:
The fix is now available in #98

PR #98, opened 2016-12-19
Changing logging option in help section

ramkrsna:
This closes the #97 and changes have been in the documentation to handle the developer intended behaviour of the script to use a `--log-file ` option instead of the` --log` option

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/98)


raghavendra-talur:
LGTM

PR #112, opened 2016-12-22
Updates for heketi autoconfig

jarrpa:
The PR https://github.com/heketi/heketi/pull/622 has caused a need to update our deployment. As heketi now uses the heketi-service-account Service Account by default and grabs all its environmental context from that, we no longer have to provide namespace, kubeapi, or even the Service Account secret.

A side-effect of this is that the OpenShift deployment now seems to require that heketi-service-account be priviledged, that is we seem to need:

```
oadm add-scc-to-user privileged system:serviceaccount:<project namespace>:heketi-service-account
```

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/112)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -459,11 +459,7 @@ kubeapi=$(${CLI} config view -o jsonpath="{.clusters[?(@.name==\"$cluster\")].cl
 
 if [[ ${LOAD} -eq 0 ]]; then
   if [[ "${CLI}" == *oc* ]]; then
-    ${CLI} process deploy-heketi -v \
-           HEKETI_KUBE_NAMESPACE=${NAMESPACE} \
-           HEKETI_KUBE_INSECURE=y \
-           HEKETI_KUBE_SECRETNAME="${heketi_secret}" \
-           HEKETI_KUBE_APIHOST="${kubeapi}" | ${CLI} create -f -
+    ${CLI} process deploy-heketi | ${CLI} create -f -
   else
     ${CLI} create -f ${TEMPLATES}/deploy-heketi-deployment.yaml
   fi
@@ -524,11 +520,7 @@ check_pods "job-name=heketi-storage-copy-job" "Completed"
 ${CLI} delete all,service,jobs,deployment,secret --selector="deploy-heketi"
 
 if [[ "${CLI}" == *oc* ]]; then
-  ${CLI} process heketi -v \
-         HEKETI_KUBE_NAMESPACE=${NAMESPACE} \
-         HEKETI_KUBE_INSECURE=y \
-         HEKETI_KUBE_SECRETNAME="${heketi_secret}" \
-         HEKETI_KUBE_APIHOST="${kubeapi}" | ${CLI} create -f -
+  ${CLI} process heketi | ${CLI} create -f -
 else
   ${CLI} create -f ${TEMPLATES}/heketi-deployment.yaml
 fi
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -465,9 +465,7 @@ if [[ ${LOAD} -eq 0 ]]; then
            HEKETI_KUBE_SECRETNAME="${heketi_secret}" \
            HEKETI_KUBE_APIHOST="${kubeapi}" | ${CLI} create -f -
   else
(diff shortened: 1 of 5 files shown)


jarrpa:
@humblec @raghavendra-talur @MohamedAshiqrh Please take a look at this. The upstream is broken without this.

jarrpa:
I'll be merging this tomorrow if no one has a look at it by then.

humblec:
@jarrpa sorry for the delay.  I feel this is kind of a change which need to be looked at in depth. I will start looking at this soon.

jarrpa:
@humblec Status? The upstream is still broken, and now has been for weeks.

humblec:
@jarrpa `A side-effect of this is that the OpenShift deployment now seems to require that heketi-service-account be priviledged,` -> What is exactly broken if the service account is not privileged ?

humblec:
@jarrpa also can you please share what is meant by upstream is broken without this ?

jarrpa:
@humblec 

1. More accurately, using current heketi master in OpenShift, the service account must belong to an SCC that is allowed to mount GlusterFS volumes. Currently only the privileged SCC is allowed to do this by default. `deploy-heketi` gets around this by not needing to mount the GlusterFS volume its creating via Kubernetes.
   * Thankfully this is only limited to heketi master and not heketi 3.x.
2. We've already broken backwards compatibility on multiple occasions. By the very fact that we've always been tracking `heketi/heketi:dev` we've NEVER had backwards compatibility because heketi itself makes no such guarantees.
   * We're looking at addressing this once heketi 4.0 is released, but at this time we can't lock to any released version because no version exists that meets our requirements (e.g. DaemonSets).
3. "The upstream is broken" means the project doesn't do what it claims to do. You can't succeed in deploying GlusterFS + heketi in either plain Kubernetes or OpenShift. This all comes down to how `heketi/heketi:dev` changed what it expects as requirements for kubeexec. See https://github.com/heketi/heketi/pull/622 .

humblec:
@jarrpa if I got it correctly the main concern is about the version of heketi in use and specifically the `heketi/heketi:dev` docker image. What about having a version specific heketi docker image in this repo ( hosted under docker hub)  and references to it from this project branches ? For ex: `this repos -> v1 -> compatible with (v3 of heketi) `. master run behind heketi upstream...etc ? so that an update in readme bring the users trust this repo and keep moving based on their choice.

jarrpa:
@humblec That seems like an unwarranted burden on this project. I see no problem with taking this PR as we have so many others that have broken compatibility and wait for heketi to release version 4 before we stabilize on a fixed version of heketi. At least that feels like a less arbitrary milestone. heketi will be releasing version 4 in January, so we only have to keep chasing it for a relatively short period of time.

humblec:
@jarrpa once again I dont have problem to take the PR in. Would u like to tag the existing commit, before proceeding further ?

humblec:
@considering this is upstream and getting stabilized, we can pull that patch in and experiemnt :).. So LGTM.

jarrpa:
@humblec The biggest one thus far has been the move to DaemonSets, that fully breaks compatibility with all released versions of heketi as none of them use or support DaemonSets for GlusterFS pods. There are other smaller changes made here and there, but the DaemonSets change is the biggest one.

And I'm happy to tag stuff, but what in particular are you looking for in a tag? :)

obnoxxx:
The problem is this:

* We use an official heketi docker image. Hence this is based on some official upstream heketi version or master.
* There is currently no upstream heketi version that works with the gk-deploy script:
  * the daemonset support is only in master.
  * the autoconfig stuff of master is not supported with gk-deploy.

So since we don't have a version with daemonset support and without autoconfig in heketi, an we won't get one, the only thing we can do here as gluster-kubernetes, is to take these patches to adapt the gk-deploy tool for autoconfig.

Once heketi v4 is out, we should probably use this version here in gluster-kubernetes.

I will do tests with this patchset and possibly merge later, but my message is that we have no alternative to taking these patches.

humblec:
@jarrpa @obnoxxx as I said in my previous comment, LGTM from my side.  I was trying to find out more details about what is broken and also exploring more options on the compatibility between the projects.

jarrpa:
@humblec Ah, sure! I just wanted to know what you meant by "tag" in this case before merging. I'll go ahead and merge it, but please feel free to let us know what you meant. :)

PR #113, opened 2016-12-22
set builtin for handling heketi-cli failures

ramkrsna:
Using the setbuiltin to handle command line execution errors and also handle the unset variables in the  deploy script 

Pair Programmed with @raghavendra-talur

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/113)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -euo pipefail
+
 PROG="$(basename "${0}")"
 TOPOLOGY='topology.json'
 LOG_FILE=''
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -420,8 +423,8 @@ if [[ "${hello}" != "Hello from Heketi" ]]; then
   abort
 fi
 
-heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY}
-heketi-cli -s http://${heketi_service} setup-openshift-heketi-storage
+heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY} /dev/null 2>&1
```
obnoxxx: should this be 
```
... > /dev/null 2>&1
```
instead of
```
... /dev/null 2>&1
```
?

deploy/gk-deploy
```diff
@@ -420,8 +423,8 @@ if [[ "${hello}" != "Hello from Heketi" ]]; then
...
+heketi-cli -s http://${heketi_service} setup-openshift-heketi-storage /dev/null 2>&1
```
obnoxxx: same here

deploy/gk-deploy
```diff
@@ -463,4 +466,3 @@ if [[ "${hello}" != "Hello from Heketi" ]]; then
 else
   output "heketi is now running."
 fi
-
```
obnoxxx: extra change?

deploy/gk-deploy
obnoxxx: and should this be an extra patch?

deploy/gk-deploy
```diff
@@ -321,15 +324,15 @@ if [[ ${ABORT} -eq 1 ]]; then
 fi
 
 if [[ "${CLI}" == *oc* ]]; then
-  ${CLI} create -f ${TEMPLATES}/deploy-heketi-template.yaml
-  ${CLI} create -f ${TEMPLATES}/heketi-service-account.yaml
-  ${CLI} create -f ${TEMPLATES}/heketi-template.yaml
+  ${CLI} create -f ${TEMPLATES}/deploy-heketi-template.yaml  || true
```
obnoxxx: why is this  `|| true` ?
i.e. why do we want to ignore errors from the ${CLI} create commands?

ramkrsna:
@obnoxxx @humblec  I have made the requested changes. Could you review.

obnoxxx:
LGTM

jarrpa:
I'm not sure I entirely like this change. Why do we need this instead of just checking the return code of the heketi-cli commands? [This post](http://mywiki.wooledge.org/BashFAQ/105) makes a pretty convincing argument as to why we would want to be very cautious of or avoid `set -e` and use our own explicit error checking mechanisms.

I also don't see a need for `-u`, unless you're thinking of making it a rule that all variables need to at least be initialized to `""`, in which case I would want to see that the script throws out an appropriate error message when expanding an unset variable.

`-o pipefail` seems fine to me.

jarrpa:
After a good face-to-face with @raghavendra-talur, I now have a bette runderstanding of the `set` options and their intended use. We will keep an eye out for any complications brought about by `-eu`, but thus far developer testing proves them to be safe.

LGTM :)

PR #114, opened 2016-12-22
Error message to handle router failure

ramkrsna:
- Handling it with a error message. The actual router check is being moved to a  prerequisite check  in the deploy script.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/114)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -417,6 +417,7 @@ debug "OK"
 hello=$(curl http://${heketi_service}/hello)
 if [[ "${hello}" != "Hello from Heketi" ]]; then
   output "Failed to communicate with deploy-heketi service."
+  output "Please verify if the router is configured. "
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -euo pipefile
```
obnoxxx: This hunk is already present in another patch (PR #113)

obnoxxx:
LGTM

jarrpa:
I feel this is incomplete, as there is no such thing as a "router" in the Kubernetes use case best I know thus this is purely for the OpenShift use case. Unless I hear an argument otherwise, I'll submit a separate PR to fix this.

Issue #126, opened 2016-12-27
Namespace accepted Even if found subset of given name


MohamedAshiqrh:
If project is created in openshift with storage-project and namespace is given to gk-deploy as "storage" then the namespace verification passes.

PR #128, opened 2016-12-27
Namespace to check exact namespace

MohamedAshiqrh:
Closes: #126 

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -377,7 +377,7 @@ if [[ -z "$NAMESPACE" ]]; then
   fi
 fi
 
-if ${CLI} get namespaces | grep -qi ${NAMESPACE}; then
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -377,7 +377,7 @@ if [[ -z "$NAMESPACE" ]]; then
   fi
 fi
 
-if ${CLI} get namespaces | grep -qi ${NAMESPACE}; then
+if ${CLI} get namespaces | grep -qiw ${NAMESPACE}; then
```
humblec: with 'i' check is not case sensitive. Do we need to skip it ?

MohamedAshiqrh: Done :)

humblec:
@raghavendra-talur PTAL. LGTM

PR #129, opened 2016-12-27
Initiating abort should skip prerequisite checks

ramkrsna:
This fixes the abort behaviour, which was prompting for the pre requisite checks to run.

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -262,7 +262,7 @@ while [[ $# -ge 1 ]]; do
   shift
 done
 
-if [[ ${LOAD} -eq 0 ]] && [[ ${SKIP_PREREQ} -eq 0 ]]; then
+if [[ ${LOAD} -eq 0 ]] && [[ ${ABORT} -eq 0 ]] && [[ ${SKIP_PREREQ} -eq 0 ]]; then
   echo "Welcome to the deployment tool for GlusterFS on Kubernetes and OpenShift.
 
 Before getting started, this script has some requirements of the execution


deploy/gk-deploy
```diff
@@ -385,7 +386,7 @@ else
   exit 1
 fi
 
-if [[ ${ABORT} -eq 1 ]]; then
+if [[ ${ABORT} -eq 1 ]] && [[ ${SKIP_PREREQ} -eq 1]]; then
```
humblec: "-eq 1]]" should be "-eq 1 ]]"

humblec:
@MohamedAshiqrh @raghavendra-talur PTAL.

ramkrsna:
@humblec sent in the commit with the required changed. @raghavendra-talur @MohamedAshiqrh PTAL

humblec:
LGTM.

deploy/gk-deploy
```diff
@@ -233,6 +233,7 @@ while [[ $# -ge 1 ]]; do
         ;;
         -abort)
         ABORT=1
+        SKIP_PREREQ=1
```
MohamedAshiqrh: Why can't we add a abort check in SKIP_PREREQ?

MohamedAshiqrh: I just think If we use SKIP_PREREQ for something else tomo then this will impact Abort as it always uses it.

MohamedAshiqrh: [Here](https://github.com/gluster/gluster-kubernetes/blob/master/deploy/gk-deploy#L265)

humblec: @MohamedAshiqrh the SKIP_PREREQ should be used for the same purpose even later. Also I feel the flag has to be enabled as soon as the option is provided. but agree on, yes there are different ways to fix it.

MohamedAshiqrh: Agreed. LGTM. No other comments.

ramkrsna: Here different solutions are possible . I'm explicitly handling only when the --abort is used. We could test the both the approaches

humblec: @ramkrsna to make it simple, we can just add the check in as mentioned by @MohamedAshiqrh . We will revisit this change later if at all required.

humblec:
LGTM. @MohamedAshiqrh PTAL.

PR #140, opened 2017-01-04
resolve that can't modprobe dm_thin_pool module

jaohaohsuan:
I found the issue that was failed to run glusterfs-daemonset.yaml on coreos. The reason is lvm couldn't initial correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/140)


humblec:
@jaohaohsuan @jarrpa I still think, this is not something we have to do from the container. The kernel modules or the space should reside independent of the container space as kernel is just a shared entity for all the processes running in the system. My knowledgede about coreos is very limited.  However module loading is something should be performed outside of the container.  Cant we load/modprobe this module from the coreos host?

PR #144, opened 2017-01-09
Strategy added to the spec of dc

MohamedAshiqrh:
Closes: #143 

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/144)

### DIFF
--- a/deploy/ocp-templates/deploy-heketi-template.yaml
+++ b/deploy/ocp-templates/deploy-heketi-template.yaml
@@ -52,6 +52,10 @@ objects:
     replicas: 1
     selector:
       name: deploy-heketi
+    triggers:
+    - type: ConfigChange
+    strategy:
+      type: Replace
     template:
       metadata:
         name: deploy-heketi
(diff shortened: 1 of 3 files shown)


humblec:
LGTM.

jarrpa:
@humblec @MohamedAshiqrh This path seems both jumbled and incomplete. The pull policy changes should have been their own commit. Additionally:
1. Shouldn't deploy-heketi also have a pull policy of IfNotPresent?
2. Shouldn't the non-Template changes go to the Kubernetes deployments as well?

MohamedAshiqrh:
@jarrpa Yeah Commit should have been more explanatory. Deploy-heketi is actually a one time thing, So not worrying much about its pull policy. Also If someone is trying deploy-heketi then it should be fresh setup is what I assume. We should also have the kube changes as I don't have a setup right now I didn't send the change. Without fail will send a pr for kube as soon as I have a setup.

jarrpa:
@MohamedAshiqrh There may be scenarios where someone wants to pre-pull the heketi image even before running gk-deploy. We actually do that [here](https://github.com/gluster/gluster-kubernetes/blob/master/vagrant/roles/nodes/tasks/main.yml#L20). :)

PR #154, opened 2017-01-12
Label cleanup

jarrpa:
Cleanup the labels given to our Kubernetes/OpenShift resources, and make sure we only delete the resources we intend to delete.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/154)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -129,18 +129,21 @@ eval_output() {
   )
 }
 abort() {
-  ${CLI} delete dc,svc,routes heketi
+  ${CLI} delete svc heketi
   ${CLI} delete sa heketi-service-account
-  ${CLI} delete secret heketi-db-backup
   if [[ "${CLI}" == *oc\ * ]]; then
+    ${CLI} delete dc,route heketi
     ${CLI} delete template heketi
+  else
+    ${CLI} delete deployment heketi
   fi
+  ${CLI} delete secret heketi-db-backup
   if [[ ${GLUSTER} -eq 1 ]]; then
     while read -r node; do
       debug "Removing label from '${node}' as a GlusterFS node."
       ${CLI} label nodes ${node} storagenode-
     done <<< "$(echo -e "${NODES}")"
-    ${CLI} delete all,service,jobs,ds,secret --selector="glusterfs"
+    ${CLI} delete ds --selector="glusterfs"
     if [[ "${CLI}" == *oc\ * ]]; then
       ${CLI} delete template glusterfs
     fi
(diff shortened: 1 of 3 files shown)


jarrpa:
@gluster/gluster-kubernetes-maintainers Please review

obnoxxx:
The second patch does some kind of partial undoing or redoing of the ${CLI} delete .... lines. Should they be squashed or somehow split up diferently?

jarrpa:
The changes to the `${CLI} delete ...` lines are to reflect the changes to the resource labels. Splitting up the commit to have the delete changes come later would leave the initial label cleanup commit in a non-working state.

obnoxxx:
sorry for being a laggard. I should have merged this after your last comment.
Now there are conflicts...

jarrpa:
@obnoxxx Yup... patches updated.

obnoxxx:
After latest changes, this needs to be updated yet again @jarrpa ...

jarrpa:
@humblec @MohamedAshiqrh Updated, please review. Note that this PR also takes care of making sure we remove the GlusterFS pods on abort.

deploy/gk-deploy
```diff
@@ -129,18 +129,20 @@ eval_output() {
   )
 }
 abort() {
-  ${CLI} delete dc,svc,routes heketi
+  ${CLI} delete all,service,deployment,secret --selector="heketi"
+  ${CLI} delete svc heketi-storage-endpoints
```
obnoxxx: shouldn't we be consistent in "svc" vs "service" ?

deploy/gk-deploy
```diff
@@ -129,7 +129,7 @@ eval_output() {
   )
 }
 abort() {
-  ${CLI} delete all,service,deployment,secret --selector="heketi"
+  ${CLI} delete service,deployment,pod --selector="heketi"
```
obnoxxx: Isn't that a semantic change: before we deleted a secret labeled heketi, now we're deleting a pod labeled heketi.

And why is "all" removed here?

jarrpa: Yes, because I'm relabeling the pod with the "heketi" label.

"all" is removed because we don't have any resources labeled "heketi" to delete that aren't service, deployments, or pods.

But I'm seeing some other changes that can be made, so I'll resubmit.

obnoxxx: but what about the secret that was there before?

jarrpa:
@obnoxxx Updated to make the PR smaller in scope.

obnoxxx:
LGTM, thanks for the updates!

Issue #149, opened 2017-01-11
Migrate project towards an ansible deployment


ravishivt:
I'm moving the ansible portion of #88 into its own issue.  I still think this project would benefit from embracing ansible as the primary deployment process.  I've started this process in my fork, https://github.com/ravishivt/gluster-kubernetes/tree/myenv/deploy.  It's not ready for a PR but I'd like to get a discussion going on if this is wanted or not before continuing work.

Here is how I use ansible in my fork, which outlines the benefits of this approach:

1. I first get a working k8s cluster up and running through whichever way I want.  IMO, a working OCP/k8s cluster should be a requirement to using this project although we can still provide a vagrant deployment or point users to other projects that focus on initial setup.
2. I then create an inventory.ini file and a topology.json file.  These must still be created manually by the end user.  The user can also edit group_vars/all.yml to modify deployment defaults.
3. I run `ansible-playbook -i inventory.ini deploy.yml`.  The deploy.yml playbook is broken down into individual ansible roles that do the following:
    1. upgrade-os: Upgrade OS packages.
    2. build-heketi: Builds heketi from source since we need heketi-cli on the master node. This may not be necessary when heketi posts latest release binaries, see heketi/heketi#627, although it is still useful to build for targets with uncommon architectures.
    3. gk-prep: Installs glusterfs-client 3.8 and loads required kernel modules.
    4. gk-deploy: Copies serval files including the gk-deploy script, k8s/ocp templates, and topology file.  It then runs the gk-deploy script.
    5. gk-post: Automatically finds the HEKETI_SERVER_CLI and creates the StorageClass under a user-configured name from group_vars/all.yml.
4. If this was a test deployment, I can easily back out by doing `ansible-playbook -i inventory.ini abort.yml destroy_vg.yml`.  This runs two playbooks that do the following:
    1. abort.yml: Runs gk-deploy with --abort.  It also removes the StorageClass.
    2. destroy_vg.yml: It lists the gluster volume groups on each node and then after user confirmation prompt, it goes ahead and deletes them using `vgremove`.

jarrpa:
I was literally JUST writing a respond to #88 as you posted this! Now I have to split it up, darn you. ;)

A few comments:

 * I'm not sure I want to force users to upgrade their entire OS. We should be checking for dependencies, checking version requirements, and then installing them where safe and appropriate. It would not be appropriate, for example, to force an upgrade of a dependency that is already installed as this could break things we're not aware of.
[...]

PR #155, opened 2017-01-14
Add a streamlined deploy option through ansible

ravishivt:
This implements #149.  It creates a new ansible/ directory with ansible playbooks and roles to deploy the project to a pre-existing cluster.  The user needs to still specify the ansible inventory file and the topology.json file.  The new ansible option does not supersede the existing manual deployment procedure or documentation.  An initial README is added with details on what the playbooks and roles do and how to use them.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/155)


jarrpa:
I want to do a careful review of this and test it out myself, so it may take me a bit. But I like it thus far!

jarrpa:
@ravishivt A bit of a non-update, since I don't want you to think I'm not thinking about this:

I've been prepping for a cross-continental trip to two conferences, so my apologies for the continued delay. I'll have plenty of time on planes and a week between the conferences to try and check this out, so I hope to get some comments back to you soon.

ravishivt:
@jarrpa No worries.  Take your time and safe travels.  One thing to note is that your ansible version must be < 2.2.x.  Ansible 2.x and after has a bug that our playbook will hit.

jarrpa:
@ravishivt Heyo. Coming in almost a month later to say that i still haven't been able to take this for a proper test run. :( My apologies. I've been hit with deadlines downstream, but we're hoping to clear them by next week.

I have, however, been able to at least read through most of it and offhand it looks pretty solid on the Debian side. I'm currently not planning to run the Debian side through its paces, mostly because I'd rather save the time of figuring out how to bring up an Ubuntu cluster (it's been long enough as it is!).

One thing to note is that with the release of heketi version 4 I plan on going off of that version for this project rather than tracking heketi master. That work will go in to the repo before this PR does, which means I'd like this playbook to build heketi v4 rather than master. I'll update this PR when that happens, and if I'm a decent human being maybe even provide patches. :)

Again, my apologies, we really do appreciate your contributions and don't like how long we've been letting this sit idle. Hope to bring some good news soon.

jarrpa:
@ravishivt All right, I've taken a crack at this and have some questions/comments:

1. Is there a reason I'm not seeing as to why the bootstrap and gk-prep roles couldn't be merged?
2. We should be able to specify cluster nodes and devices in the inventory.ini file and just generate a topology file.
3. I know that for `yum` at least if you specify a package that is already installed it will still succeed, just with a warning. Is `apt-get` similar? If so we don't need to check for python already existing.
4. heketi 4 has been released, so we should build that version

goern:
I wonder if you could provide a heketi-cli container as I cant build it on an Fedora Atomic Host

obnoxxx:
@goern how does that request/question relate to this PR?

It would certainly be possible. up to now we've never needed to build a specific cli container.

jarrpa:
Closing this PR. It can be reopened if needed.

PR #156, opened 2017-01-17
Use heketi's service IP and port

ravishivt:
Updated documentation to fetch the heketi service IP and port instead of the service's endpoint IP and port.  When users follow the docs (base README and hello world README), they set up a storage class that is set up to talk to the heketi pod directly.  This will work but will fail if the pod gets deleted and recreated with a new IP.

The only disadvantage of this PR's approach is that the `jq` package must be installed.  We could look at other one-liners to accomplish this without 3rd party packages but `jq` makes it pretty clean.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/156)

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ to go. You can set the `HEKETI_CLI_SERVER` environment variable as follows so
 that it can be read directly by `heketi-cli` or sent to something like `curl`:
 
 ```bash
-$ export HEKETI_CLI_SERVER=$(kubectl describe svc/heketi | grep "Endpoints:" | awk '{print "http://"$2}')
+$ export HEKETI_CLI_SERVER=$(kubectl get svc/heketi --template 'http://{{.spec.clusterIP}}:{{(index .spec.ports 0).port}}')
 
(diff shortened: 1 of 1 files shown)


ravishivt:
Ignore the part about `jq` being a requirement.  I learned that kubectl has a `--template` option that takes a GO template.  PR is updated.

jarrpa:
Also, yes! Good catch, I was thinking there was something about go templates we could make use of. ;) Can you squash the final patches into one? We don't need to commit the step we took through `jq`.

ravishivt:
It doesn't add the `http://`.  I will add it and squash the commits.

ravishivt:
squashed

jarrpa:
Awesome! :lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/gluster/gluster-kubernetes/156#-:-KanDeHr1JV9fl1G1q7l:bfgc9op)*

PR #169, opened 2017-02-17
Removing kubectl CLI option, making oc as default

ramkrsna:
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/169)


jarrpa:
@ramkrsna #170 seems to be the same general work, though this one has a formatting error, and has an earlier date. I'll be taking that one instead.

PR #170, opened 2017-02-17
gk-deploy: default to 'oc' as cli if found.

obnoxxx:
The assumption is that in a OpenShift environment, users want
to use oc. They can still force the use of kubectl with the --cli
switch.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/170)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -329,23 +329,7 @@ fi
 if [[ "x${CLI}" == "x" ]]; then
   kubectl=$(type kubectl 2>/dev/null | awk '{print $3}')
   oc=$(type oc 2>/dev/null | awk '{print $3}')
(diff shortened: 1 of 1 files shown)


obnoxxx:
@jarrpa What do you think?

I think it is confusing to always have to reply to this when the answer is almost always  "oc". And we can still override the cli to kubectl via cmdline.

obnoxxx:
Oops, was not aware that @ramkrsna hat proposed a patch today as #169.

jarrpa:
@obnoxxx I would rather have it so that the prompt can be skipped with `-y` while defaulting to `oc` instead, but I can see the argument for this change. I'll take it now, and if anyone raises an issue about it we can revert and do it with `-y` instead. :)

PR #175, opened 2017-02-27
check existing setup before deployment

raghavendra-talur:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/175)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -155,14 +155,18 @@ assign() {
   keypos=$keylen
 }
 
-check_pods() {
-  local all_ready=false
+are_pods_running() {
+  local all_ready="no"
+  local wait_for=2
   s=0
+  if [ -n "$2" ]; then
+    wait_for="$2"
+  fi
   debug "\nChecking status of pods matching '${1}':"
-  while [[ "${all_ready}" != "true" ]]; do
-    if [[ ${s} -ge ${WAIT} ]]; then
+  while [[ "${all_ready}" != "yes" ]]; do
+    if [[ ${s} -ge $wait_for ]]; then
       output "Timed out waiting for pods matching '${1}'."
-      abort
+      break
     fi
     sleep 2
     pods=$(${CLI} get pod --no-headers --show-all --selector="${1}" 2>&1)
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -384,6 +384,11 @@ if [[ ${ABORT} -eq 1 ]]; then
 fi
 
 if [[ ${LOAD} -eq 0 ]]; then
+  check_pods "glusterfs=heketi-pod"
```
MohamedAshiqrh: @raghavendra-talur My Bad. Check pods will abort if it didn't find the pods with these labels. 
I think we can add another flag which can parse in check pods and return appropriate. It is upto you.

raghavendra-talur: Done. Thanks.

deploy/gk-deploy
```diff
@@ -384,6 +389,10 @@ if [[ ${ABORT} -eq 1 ]]; then
 fi
 
 if [[ ${LOAD} -eq 0 ]]; then
+  if [[ $(are_pods_running "glusterfs=heketi-pod") == "yes" ]]; then
```
MohamedAshiqrh: We can use True or false?

MohamedAshiqrh: We can ignore for now.

MohamedAshiqrh:
@humblec review and merge.

humblec:
LGTM.

jarrpa:
Why are we even modifying check_pods to return a string? If I don't hear back on this by tomorrow I'm reverting it.

raghavendra-talur:
@jarrpa Earlier, check_pods was not returning anything. If pods are found, control flows back to caller and it proceeds as everything is good. If pods are not found, check_pods called abort() which lead to an exit.

IMO, check_pods is a util function which not change control flow of the program. Hence the change to make it return value to indicate if pods are running or not. The action to be taken is left to the caller.

If the question is why "string" instead of a int, it is just a preference that I have. We can change it to int or bool if you strongly feel so.

jarrpa:
@raghavendra-talur Okay, agreed on the control flow point. However in testing I found other concerns, in particular that output is no longer properly displayed when running the command in a subshell. I'm submitting an alternative implementation now.

PR #181, opened 2017-02-27
check_pods() return code

jarrpa:
Implement and check a return code from check_pods(). Also allow a custom wait timer for a given check_pods() call.

These two commits are in the same PR because they both rely on the revert in the first commit. These are an alternative implementation to #175 due to some issues encountered in testing.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/181)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -156,14 +156,14 @@ assign() {
 }
 
 check_pods() {
-  local all_ready=false
+  local rc=1
   local wait_limit=${2:-${WAIT}}
   s=0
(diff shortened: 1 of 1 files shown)


obnoxxx:
While I don't like reverts very much, and the original patch has a slight advantage of the calling scheme (if [[ $(are_pods_running "glusterfs=heketi-pod") == "yes" ]]; ...) in being slightly more idiomatic, this proposed change is overall more consistent, imho, e.g. it's better to use real RC for tracking result, and it does not sneak in a function name change under the hood ;-)...  Hence LGTM.

PR #178, opened 2017-02-27
Print all output to log file

MohamedAshiqrh:
Signed-off-by: Mohamed Ashiq Liyazudeen <mliyazud@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/178)

PR #182, opened 2017-02-27
Properly capture subcommand output in both stdout and log file

jarrpa:
A modification to  #178 that allows for subcommand output to be sent to stdout (and the log file) in real time. This also allows the capture of any return codes from the subcommands.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/182)


obnoxxx:
Neat! LGTM

PR #183, opened 2017-02-28
Fix regression in topology load error reporting

jarrpa:
Allow the script to detect if a device or node has failed to load
from the topology file and alert the user on what to do. This is
fixing a regression introduced by PR #182.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/183)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -479,7 +479,20 @@ else
   output "heketi is now running."
 fi
 
-eval_output "heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY} 2>&1"
-if [[ $? -ne 0 ]]; then
-  output "Please check the failed node or device and restart with --load"
+load_temp=$(mktemp)
+eval_output "heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY} 2>&1" | tee ${load_temp}
+grep -q "Unable" ${load_temp}
+unable=$?
+rm ${load_temp}
+
+if [[ ${PIPESTATUS[0]} -ne 0 ]] || [[ ${unable} -eq 0 ]]; then
+  output "Error loading the cluster topology."
+  if [[ ${unable} -eq 0 ]]; then
+    output "Please check the failed node or device and rerun this script using the --load option."
+  fi
+  exit 1
+else
+  output "heketi topology loaded."
 fi
+
+output "Ready to create and provide GlusterFS volumes."


deploy/gk-deploy
```diff
@@ -479,7 +479,20 @@ else
...
-eval_output "heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY} 2>&1"
-if [[ $? -ne 0 ]]; then
-  output "Please check the failed node or device and restart with --load"
+load_temp=$(mktemp)
+eval_output "heketi-cli -s http://${heketi_service} topology load --json=${TOPOLOGY} 2>&1" | tee ${load_temp}
+grep -q "Unable" ${load_temp}
+unable=$?
+rm ${load_temp}
+
+if [[ ${PIPESTATUS[0]} -ne 0 ]] || [[ ${unable} -eq 0 ]]; then
```
MohamedAshiqrh: Just curious what does PIPESTATUS do?

jarrpa: It's one of those sneaky internal BASH variables. :) From the [Advanced Bash-Scripting Guide](http://tldp.org/LDP/abs/html/internalvariables.html):

> $PIPESTATUS
>     Array variable holding exit status(es) of last executed foreground pipe.
> 
> ```
>     bash$ echo $PIPESTATUS
>     0
> 
>     bash$ ls -al | bogus_command
>     bash: bogus_command: command not found
>     bash$ echo ${PIPESTATUS[1]}
>     127
> 
>     bash$ ls -al | bogus_command
>     bash: bogus_command: command not found
>     bash$ echo $?
>     127
> ```
>The members of the $PIPESTATUS array hold the exit status of each respective command executed in a pipe. $PIPESTATUS[0] holds the exit status of the first command in the pipe, $PIPESTATUS[1] the exit status of the second command, and so on.

MohamedAshiqrh: Awesome. thanks.

deploy/gk-deploy
```diff
@@ -479,7 +479,20 @@ else
...
+  output "Error loading the cluster topology."
+  if [[ ${unable} -eq 0 ]]; then
+    output "Please check the failed node or device and rerun this script using the --load option."
+  fi
+  exit 1
+else
+  output "heketi topology loaded."
 fi
+
+output "Ready to create and provide GlusterFS volumes."
```
obnoxxx: Don't sneak in stuff that does not actually belong to the commit...

I'll be generous this time, but less so in the future ;-)

obnoxxx:
LGTM

PR #184, opened 2017-02-28
[WIP] Unit Testing: Initial framework

MohamedAshiqrh:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/184)

### DIFF
--- a/deploy/tests/test.sh
+++ b/deploy/tests/test.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+# Copyright (c) 2016 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+KUBE_GLUSTER_DEAMONSET=../kube-templates/glusterfs-daemonset.yaml 
+KUBE_HEKETI_DEPLOYMENT=../kube-templates/heketi-deployment.yaml
+KUBE_HEKETI_SA=../kube-templates/heketi-service-account.yaml 
+OCP_GLUSTER_TEMPLATE=../ocp-templates/glusterfs-template.yaml
+OCP_HEKETI_TEMPLATE=../ocp-templates/heketi-template.yaml
+OCP_HEKETI_SA=../ocp-templates/heketi-service-account.yaml
+TOPOLOGY=../topology.json.sample
+
+# syntax checking can not be done if file has more than one yaml context 
+# so split and call check
+# check kube-templates/heketi-deployment.yaml to understand what I am talking about
+split_yaml_check() {
+  file="${1}"
+  awk '{print $0 > "file" NR}' RS='---'  ${file}
+  count=$(ls -l file* | wc -l)
+  for i in `seq 1 ${count}`
+  do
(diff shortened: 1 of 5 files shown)


MohamedAshiqrh:
@jarrpa @obnoxxx @humblec @raghavendra-talur @ramkrsna Please Review.

MohamedAshiqrh:
@raghavendra-talur Nice points. Worked like charm. Thanks.

MohamedAshiqrh:
Now it Just tests for yaml syntax err. Mock func to be added and gk-deploy-unit should be used for testing.

MohamedAshiqrh:
Still 100 more lines to go. Have to add more validations and exit conditions. validation for example, We don't validate the Template file locations specified or the defaults. 
@jarrpa Should I add the changes to gk-deploy along with this PR or send a new PR?

jarrpa:
@MohamedAshiqrh That should be a separate PR. :)

deploy/tests/test.sh
```diff
@@ -0,0 +1,97 @@
...
+  file="${1}"
+  TEST=$(python -c 'import yaml,sys;yaml.safe_load(sys.stdin)' < ${file} 2>/dev/null)
+  if [[ $? -ne 0 ]]; then
+    echo "Failed ${file}"
+  fi
+}
+
+# copies the source and adds the mock source. This will be used for testing
+cp ../gk-deploy gk-deploy-unit
+sed -i '16s/^/\nsource mock.sh/' gk-deploy-unit
```
obnoxxx: This works, but I don't like this copying and modifying of the script very much.

It should be sufficient to source the mock.sh in *THIS* script, right?
If I am mistaken, could we possibly use a checked-in wrapper script?

deploy/tests/test.sh
```diff
@@ -0,0 +1,97 @@
...
+
+# call all the yaml files and check if they exist. Also call yaml syntax checking
+for i in $KUBE_GLUSTER_DEAMONSET $KUBE_HEKETI_DEPLOYMENT $KUBE_HEKETI_SA $OCP_GLUSTER_TEMPLATE $OCP_HEKETI_TEMPLATE $OCP_HEKETI_SA 
+do
+  if test "$(ls ${i})"
+  then
+	echo "$i is present"
+	# Add here if the new file has more than one yaml context
+        if [[ "${i}" != "${KUBE_HEKETI_DEPLOYMENT}" ]]; then
+	  check "${i}"
```
obnoxxx: indentation

deploy/tests/test.sh
```diff
@@ -0,0 +1,97 @@
...
+KUBE_HEKETI_SA=../kube-templates/heketi-service-account.yaml 
+OCP_GLUSTER_TEMPLATE=../ocp-templates/glusterfs-template.yaml
+OCP_HEKETI_TEMPLATE=../ocp-templates/heketi-template.yaml
+OCP_HEKETI_SA=../ocp-templates/heketi-service-account.yaml
+TOPOLOGY=../topology.json.sample
+
+# syntax checking can not be done if file has more than one yaml context 
+# so split and call check
+# check kube-templates/heketi-deployment.yaml to understand what I am talking about
+split_yaml_check() {
```
obnoxxx: It should be possible to write this function so that the case of a single context (even without opening "---\n") would be a valid corner case. So the caller should not need to know and just call this function, ideally under a different name, like yaml_syntax_check(...) or so...

deploy/tests/test.sh
obnoxxx: I tested it, see https://github.com/obnoxxx/gluster-kubernetes/tree/pr-184-plus . (This also contains a few cleanups in the formatting etc).

But there seem to be issues. the sourcing in the test script does not work immediately.

I noticed a problem: When you just comment out the creation of the patched script, the test runs a little longer but succeeds nonetheless even though no kubectl or oc commands are available.. :-o This needs to be fixed.

obnoxxx:
This was superseded by #194 which was based on this ideas.

PR #186, opened 2017-03-02
Changes to generic gk-deploy functions

jarrpa:
This PR moves some generic functions within gk-deploy into a separate file, gk-functions, so that they can be consumed by other scripts as well. This also includes a few minor fixes to the assign() function. I didn't want to split them since they are inter-dependent.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/186)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+. ./gk-functions
+
 PROG="$(basename "${0}")"
 TOPOLOGY='topology.json'
 LOG_FILE=''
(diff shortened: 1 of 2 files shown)


humblec:
@MohamedAshiqrh PTAL.

deploy/gk-deploy
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+. ./gk-functions
```
MohamedAshiqrh: why can't we do source gk-functions? I mean I just came to know about source when I tried the unit testing part. I just wanna know is there any advantage in this/source.

deploy/gk-deploy
```diff
@@ -148,25 +113,6 @@ abort() {
...
-    return 2
-  else
-    output "Required parameter for '-${key}' not specified.\n"
-    usage
-    exit 1
-  fi
-  keypos=$keylen
-}
-
 check_pods() {
```
MohamedAshiqrh: Can we have check pods separate too?

deploy/gk-deploy
jarrpa: `.` is a synonym for `source` :)

deploy/gk-deploy
jarrpa: It depends on having $CLI defined, so I'd say no. Plus I haven't found a need for it yet.

obnoxxx:
Generally looks like a clean change, even though it means we can't simply copy around a single script file any more. (but well... we need the templates along with it anyways...)

@jarrpa Could you also add a comment what "assign()" does? To me this is one of those totally opaque shell magic functions, where I can't easily tell what it actually does from looking at it... :-)

jarrpa:
@obnoxxx Updated to match current versions of the functions in master. Best I've check I don't lose any functionality, but please verify, especially with assign().

Also added documentation. :)

I will note: with the move to officially use openshift-ansible for the registry stuff, the immediate need for this PR is greatly reduced. I'm more than willing to consider closing this PR until actually needed.

obnoxxx:
@jarrpa will take a closer look.
But what's the original motivation: I.e. are we expecting more scripts to use these?

jarrpa:
@obnoxxx Originally I was going to implement the registry integration here, and I was going to have a separate gk-registry script that used these functions. At this time that will not be happening, and we have no other plans for scripts to come in. Hence why we may not need this.

jarrpa:
Closing this PR for now.

PR #190, opened 2017-03-08
Freeze on stable versions of component projects

jarrpa:
Use images based on stable releases of our component projects rather than using images based on their master branches. Also update the README to document these versions.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/190)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,9 @@ Kubernetes.
 
 ### Component Projects
 
-* **[Kubernetes](http://kubernetes.io/)**, the container management system.
-* **[GlusterFS](https://www.gluster.org/)**, the scale-out storage system.
-* **[heketi](https://github.com/heketi/heketi)**, the RESTful volume management
+* **[Kubernetes](http://kubernetes.io/) 1.5**, the container management system.
+* **[GlusterFS](https://www.gluster.org/) 3.9**, the scale-out storage system.
+* **[heketi](https://github.com/heketi/heketi) 4.0**, the RESTful volume management
   interface for GlusterFS.
 
 ### Presentations
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -19,7 +19,7 @@ spec:
         storagenode: glusterfs
       hostNetwork: true
       containers:
-      - image: gluster/gluster-centos:latest
+      - image: gluster/gluster-centos:gluster3u9_centos7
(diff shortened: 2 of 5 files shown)


obnoxxx:
We need to decide where to apply this patch:

* do we want to put master to released versions or keep it at "dev" versions?
* If we want to put master to released versions, then we can continue to do our current model of releasing directly off of master.
* if instead we want to leave master at "dev" versions, we need to create branches.
  *  Either for each release create a release branch when we start a new release and apply this patch as the very first thing.
  * Or keep a parallel "stable" branch with this patch applied. keep picking patches from master to stable (if applicable) and then create releases off of the stable branch.

None of these approaches looks entirely convincing to me ...

jarrpa:
OH right, it would help to record some thoughts that were had in other channels.

If we number the options list you presented, I'm all for option 3. I think there is something to be gained by having by having the main branch be the latest bleeding edge tech (that we're willing to tolerate, anyway). As for how we go about structuring the other branches, I lean towards having one branch per major release, mostly because it's familiar to me. I can see having a single stable branch working as well, but we'd probably want to do cherry-picks for minor releases and then merges with master for major releases... and I hate merges. ;P

jarrpa:
@obnoxxx Updated to current master.

@MohamedAshiqrh @humblec @ramkrsna @raghavendra-talur We'd like someone else to weigh in on the conversation. Currently the two approaches I'm in favor of look as follows:

Separate release branches with cherry-picks and tags for releases (aka the Samba model). This leaves the git history with fewer merges, making it easier to do `git bisect`. This has the downside of making it slightly less obvious how to find the latest stable version:
```
G  R2.1 (G`)
|  |
F  F`
|  |
|  R2.0 (E`)
| /
E
|
D  R1.1 (D`)
|  |
C  |
|  |
|  R1.0 (B`)
| /
B
|
A
```
Single stable branch with cherry-picks for minor releases, merges for major releases, and tags. While it has more merges, this would make it simple to check out the stable branch for the latest stable release.
```
G  R2.1 (G`)
|  |
F  F`
|  |
|  R2.0
| /|
E  |
|  |
D  R1.1 (D`)
|  |
C  |
|  |
|  R1.0
| /
B
|
A
```
As mentioned, I am personally leaning towards the first method, in particular because I like the cleaner git history. However I can see the merits of the second method.

Let us know! :)

jarrpa:
Closing this for now, in favor of something coming from #272.

PR #207, opened 2017-03-21
Adding a simple containerized heketi and dedicated gluster example

screeley44:
This is a simple single node Kube cluster example

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/207)

### DIFF
--- a/docs/examples/containerized_heketi_dedicated_gluster/README.md
+++ b/docs/examples/containerized_heketi_dedicated_gluster/README.md
@@ -1,0 +1,697 @@
+# Containerized Heketi with an dedicated GlusterFS cluster (static and dynamic provisioning)
+
+Using GlusterFS and Heketi is a great way to perform static and dynamic provisioning 
+in a Kubernetes based cluster for persistent volume shared filesystems.  Typically this is accomplished by running both
+Heketi and GlusterFS as containerized components of the cluster (gk-deploy), but what about only running Heketi as a container 
+and communicating with a dedicated GlusterFS cluster (non-containerized)?
+
+This example will walk through a simple version of this setup to take advantage of existing dedicated storage environments while
+still providing the benefits of reliability and failover from a Kubernetes environment.
+
+This example assumes some familiarity with Kubernetes and the [Kubernetes Persistent Storage](http://kubernetes.io/docs/user-guide/persistent-volumes/) model.
+
+This example also assumes you have access to an existing GlusterFS cluster that has raw devices available for consumption and management by Heketi Server.  If 
+you do not have this, you can create a 3 node (CentOS or fedora) cluster using your Virtual Machine solution of choice and make sure you create a few raw devices and give
+plenty of space.  See below for a quick start guide or [installing GluserFS Guide](https://www.gluster.org/)
+
+
+### Environment and prerequisites for this example.
+
+- GlusterFS Cluster running CentOS 7 (2 nodes, each with at least 2 X 200GB raw devices)
+   - gluster25.rhs (192.168.1.205)
+   - gluster26.rhs (192.168.1.206)
+
+- Kubernetes Node running CentOS 7 (1.5 or later, this example will use the latest source from HEAD for 1.7 version)
+   - k8dev3.rhs (192.168.1.209)
+   - run a single node cluster using ./hack/local-up-cluster.sh
+
+***NOTE:*** When running local-up-cluster.sh to build a single node Kubernets cluster I typically run export HOSTNAME_OVERRIDE=yourhostname
+
+### Verify functioning and accessible dedicated GlusterFS environment
+
+Below are some basic commands to help validate that your GlusterFS cluster is working properly.
+Also, for this example, an existing volume was created to store the Heketi db file, to simulate a dedicated environment
+with existing volumes.
+
+#### Installing and configuring GlusterFS Nodes (If needed)
+
+Install 2 nodes to run as the GlusterFS storage cluster and simulate an existing legacy environment. 
+This example uses 2 CentOS 7 systems all running latest GlusterFS. More advanced and additional information can be found on the official [GlusterFS Site](https://www.gluster.org/)
+
+***NOTE:*** 2 node clusters are not officially recommended as it causes some known replication issues (split brain), and if possible a 3+ node cluster should be used if possible.
+
+#### View basic node and device information.
+
+Execute `lsblk` command to show our raw devices (which will be used later in this example to load into Heketi).
+
+```
+	# lsblk
+	NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
+	sda           8:0    0   20G  0 disk 
+	├─sda1        8:1    0    1G  0 part /boot
+	└─sda2        8:2    0   19G  0 part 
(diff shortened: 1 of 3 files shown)


screeley44:
@jarrpa @humblec - PTAL

docs/examples/containerized_heketi_dedicated_gluster/README.md
```diff
@@ -0,0 +1,669 @@
...
+  - name: heketi
+    port: 8888
+    targetPort: 8888
+
+```
+
+heketi-rc.yaml
+```
+apiVersion: v1
+kind: ReplicationController
```
jarrpa: Why are you using a ReplicationController? I think I have an idea why, but I want to know your view.

docs/examples/containerized_heketi_dedicated_gluster/README.md
```diff
@@ -0,0 +1,669 @@
...
+  replicas: 1
+  selector:
+    app: heketi
+  template:
+    metadata:
+      name: heketi
+      labels:
+        app: heketi
+    spec:
+      hostNetwork: true
```
jarrpa: Why is this being used?

docs/examples/containerized_heketi_dedicated_gluster/README.md
```diff
@@ -0,0 +1,669 @@
...
+master node.
+
+
+```
+	# export HEKETI_CLI_SERVER=http://k8dev3.rhs:8888
+```
+
+### Using Heketi with Gluster
+
+Topology is used to tell Heketi about the environment, what nodes and devices it will manage.
```
jarrpa: This sentence is clunky, please reword. Start with something like "A topology file" instead.

docs/examples/containerized_heketi_dedicated_gluster/README.md
```diff
@@ -0,0 +1,669 @@
...
+          "devices": [
+            "/dev/sdc"
+          ]
+        }
+      ]
+    }
+  ]
+}
+```
+
```
jarrpa: Include a note about how "manage" should be hostnames and "storage" should be IP addresses. I can guarantee you someone will stumble over this by not having read it in the three-ish other places we have it noted. :)

docs/examples/containerized_heketi_dedicated_gluster/README.md
screeley44: hostNetwork allowed my pod to correctly access my local private network where gluster is running (192.168.x.x).  I was not able to properly get the Heketi pod to easily communicate out of the cluster and reach my gluster nodes.  I'm not sure what I'm missing, but this seemed like a reasonable approach that assume that your nodes have network access so why not have the pod take advantage of that without having to introduce more complex network concepts (dns, routing tables, etc...).  I'm open to other suggestions but just haven't had luck without it.  This also may be a side product of running a single node local cluster as well, assuming a more production like environment would have better networking support and resources.

docs/examples/containerized_heketi_dedicated_gluster/README.md
screeley44: Well, since heketi is basically stateless in this case, as we are storing the heketi.db on our actual Gluster server, I didn't see a need for this simple kick-the-tires demo to use a daemonset or deployment.  However, I could see the value in being consistent with the existing deploy yaml files, would you prefer I switch to a deploymentconfig for heketi?

jarrpa: Let's go for consistency where possible. :)

docs/examples/containerized_heketi_dedicated_gluster/README.md
jarrpa: Ah, all righty!

docs/examples/containerized_heketi_dedicated_gluster/README.md
screeley44: agreed, the more I think about it, I think I will use a daemonset for heketi, I see value in having it run on every pod actually using a loadbalancer service type...what do you think?

screeley44: @jarrpa - do you know if there is an issue with having multiple instances of heketi running in a single cluster, seems as long as heketi.db is in a  shared location and accessible by each node that there wouldn't be an issue?

screeley44:
@jarrpa - regarding

```
Is there any reason to not do the config and keys as a ConfigMap and Secret repectively? I would imagine it would be easier than making sure all nodes have the same content in the right directories.
```

I agree with the ConfigMap, and made that change, I'm a bit worried about the Secret's (although I also made that change, but my environment is very simple being it is a single node dev cluster) - I'd have to test on a more production like cluster, but if I have multiple nodes wouldn't I need a secret for each nodes keys?  At that point, I'm generating keys on each node and then copying them back to my master so I can create the ssh-key-secret with each keyfile from each node?

docs/examples/containerized_heketi_dedicated_gluster/README.md
jarrpa: I don't think multiple heketi services can access the same db. That would be the main reason we would've used a Deployment instead of a DaemonSet.

jarrpa:
@screeley44 Couldn't you just generate one key that's the heketi key, store it in a Secret, and then copy that one key to all the GlusterFS nodes? I don't think it would matter what host it's coming from if it's the same key.

screeley44:
@jarrpa - I think I addressed all your concerns, give it another once over and let me know what you think.

jarrpa:
@screeley44 Thanks! A few still go unaddressed. :)

 1. The sentence I pointed out is still clunky
 2. There is no note on the hostname/IP requirements for the topology file
 3. I still don't understand why you need to generate a key on every node that might run the heketi pod. You should only have to generate one public/private pair.
 3.1. Since the keys are in their own directory, you can just specify `--from-file=/path/to/dir` to get them both
 4. *New:* The Deployment file is named `heketi-dc.yaml`, but DCs don't exist in Kubernetes. Please rename to `heketi-deployment.yaml`.
5. *New:* The `heketi-rc.yaml` file is still in the branch

screeley44:
@jarrpa 
1 - yep, missed it, will address on next push

2 - you want the actual <1> type of note?  I had the NOTE at the bottom, maybe I'll add both, sometimes these things are hard to see

```
NOTE: The manage element in the above json file should be the actual hostname of the GlusterFS node, and the storage element should be the actual ip address of the storage node as it is in the glusterfs-endpoints.
```

3 - yep, I just missed updating that part of the doc - next push

4 - good catch

5 - another good catch will remove

thanks!

docs/examples/containerized_heketi_dedicated_gluster/README.md
obnoxxx: Is that related to the fact that heketi when running in kube uses kubeexec by default to talk to gluster, instead of sshexec?

jarrpa: Shouldn't be, he's over-writing the default behavior with his own config file.

jarrpa:
LGTM

PR #195, opened 2017-03-11
Add a vagrant/openshift environment

obnoxxx:
in parallel to the kubernetes env.
Based on vagrant-heketi.

PR #214, opened 2017-03-27
QOS: adding Memory limitaion for the gluster pod

MohamedAshiqrh:
A Guaranteed memory container gets the amount of memory requested,
but no more. In the event of an out of memory event, it will only
be killed at last on the system.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/214)

### DIFF
--- a/deploy/ocp-templates/glusterfs-template.yaml
+++ b/deploy/ocp-templates/glusterfs-template.yaml
@@ -35,6 +35,11 @@ objects:
         - image: gluster/gluster-centos:latest
           imagePullPolicy: IfNotPresent
           name: glusterfs
+          resources:
+            requests:
+              memory: 20Gi
+            limits:
+              memory: 32Gi
           volumeMounts:
           - name: glusterfs-heketi
             mountPath: "/var/lib/heketi"
@@ -83,7 +88,6 @@ objects:
             periodSeconds: 25
             successThreshold: 1
             failureThreshold: 15
-          resources: {}
           terminationMessagePath: "/dev/termination-log"
         volumes:
         - name: glusterfs-heketi
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -22,6 +22,11 @@ spec:
       - image: gluster/gluster-centos:latest
         imagePullPolicy: IfNotPresent
         name: glusterfs
+        resources:
+          requests:
+            memory: 20Gi
+          limits:
+            memory: 32Gi
         volumeMounts:
         - name: glusterfs-heketi
           mountPath: "/var/lib/heketi"


MohamedAshiqrh:
@obnoxxx @jarrpa @humblec please review the Memory limitation has to be discussed. I have just added a tested syntax.

jarrpa:
What do you mean by "at last on the system"? Also, will this allocate 32GB at the outset or just up to 32GB?

MohamedAshiqrh:
It will allocate up to 32 GB. 

https://docs.openshift.com/container-platform/3.4/dev_guide/compute_resources.html#dev-compute-resources

MohamedAshiqrh:
A **BestEffort** memory container is able to consume as much memory as is available on the node, but there are no guarantees that the scheduler will place that container on a node with enough memory to meet its needs. In addition, a BestEffort container has the greatest chance of being killed if there is an out of memory event on the node.

A **Burstable** memory container is scheduled on the node to get the amount of memory requested, but it may consume more. If there is an out of memory event on the node, Burstable containers are killed after BestEffort containers when attempting to recover memory.

A **Guaranteed** memory container gets the amount of memory requested, but no more. In the event of an out of memory event, it will only be killed if there are no more BestEffort or Burstable containers on the system.

humblec:
@jarrpa pod api has it . https://docs.openshift.com/container-platform/latest/rest_api/openshift_v1.html#v1-pod

Also, https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/ has more information on CPU and memory setting.

humblec:
@jeremyeder @ekuric Can you also review this PR ?

humblec:
@MohamedAshiqrh Bugzilla c#https://bugzilla.redhat.com/show_bug.cgi?id=1431590#c13

jarrpa:
@humblec I just did a search on that page for "Tier" and it was nowhere to be found. :) Even so, this still seems like a generated field, and no one that should be configured.

Yes, I read that. My other comments still stand. Per that link you posted:
```
spec:
  containers:
  - name: cpu-ram-demo-container
    image: gcr.io/google-samples/node-hello:1.0
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "1"
```

`requests` and `limits` do not exist under `spec`, they exist under `spec.resources`. ALSO it appears you can achieve a Guaranteed QoS Tier by only specifying a limit.

@MohamedAshiqrh Please take my changes into account, and make sure you test this.

humblec:
@jarrpa Sure. Indeed we have to test this well. Btw OoS Tiers are explained here https://docs.openshift.com/enterprise/3.1/dev_guide/compute_resources.html#quality-of-service-tiers .

jarrpa:
@humblec Yes, I read that too. :) And indeed, that's the document I'm referencing. To quote it directly:

> A **Guaranteed** quality of service is provided when a limit is specified that is equal to an optionally specified request.

So we don't need to specify a `request`.

deploy/ocp-templates/glusterfs-template.yaml
```diff
@@ -35,6 +35,12 @@ objects:
         - image: gluster/gluster-centos:latest
           imagePullPolicy: IfNotPresent
           name: glusterfs
+          QoS Tier:
+            memory: Guaranteed
```
jeremyeder: This needs to be Burstable for now due to the current state of QoS tiers in Kubernetes.  Per @derekwaynecarr comment in the BZ:

```
@jerms, i might be inclined to have the QoS class Burstable where the cpu
request is specified, but the limit is not.  A memory limit should be
specified.  That feels like the sweet spot today given the current state of the
product, WDYT?```

Comment (jeremyeder) [inline deploy/ocp-templates/glusterfs-template.yaml]: We expect this to change at some point to make Guaranteed more suitable for these types of workloads. But we think the Guaranteed tier is currently NOT what we should do.

Comment (jarrpa) [inline deploy/ocp-templates/glusterfs-template.yaml]: @jeremyeder So specify both a CPU and memory `request`, but only specify a `memory` limit?

Comment (MohamedAshiqrh) [inline deploy/ocp-templates/glusterfs-template.yaml]: Addressed the comment and made it Burstable.

**Comment by MohamedAshiqrh** [comment]
@jarrpa I am making the changes to kube deployment.

**Comment by jarrpa** [comment]
I'm still not seeing the changes I requested.

* QoS Tier is not listed as a valid field anywhere in the spec, please remove it.
* `Limits` and `Requests` do not exist under `spec.containers`, but they do exist in `spec.containers.resources`. They should probably be moved there, and should also be lowercase
  https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container
* If we're going with Burstable strategy now, (this blog post)[http://www.devoperandi.com/kubernetes-container-guarantees-and-oversubscription/] that I linked earlier indicates that Burstable is achieved by the following:
  > Burstable if request less than limit OR one of the containers has nothing set
  > ```
  > containers:
  >     name: foo
  >         resources:
  >             limits:
  >                 cpu: 10m
  >                 memory: 1Gi
  >             requests:
  >                 cpu: 10m
  >                 memory: 1Gi
  >     name: bar
  > ```
  > 
  > now recognize there are two containers above, one with nothing specified so that container gets BestEffort which makes the Pod as a whole Burstable.

Has this actually be tested to work? How can this be verified so I can see the results myself?

**Comment by obnoxxx** [comment]
@MohamedAshiqrh tests failed. does this need to be rebased and retested?
@jarrpa: any comment on the updated patch?

**Comment by MohamedAshiqrh** [comment]
rebased

**Comment by obnoxxx** [comment]
forget about rebasing, @MohamedAshiqrh, that was for the heketi repo... ;-)

**Comment by jarrpa** [comment]
@MohamedAshiqrh @obnoxxx There's no update to the kube template. :)

Also, a request of 20GB still seems rather high, but I'll forgo that judgment since I have no real experience with the memory consumption of these pods.

**Comment by MohamedAshiqrh** [comment]
@jarrpa I have added and verified the Openshift template.

`# oc describe pods glusterfs-djz9g
Name:			glusterfs-djz9g
Namespace:		block-test
Security Policy:	privileged
Node:			192.168.36.2/192.168.36.2
Start Time:		Thu, 17 Aug 2017 08:41:40 -0400
Labels:			glusterfs-node=pod
			pod-template-generation=1
Annotations:		kubernetes.io/created-by={"kind":"SerializedReference","apiVersion":"v1","reference":{"kind":"DaemonSet","namespace":"block-test","name":"glusterfs","uid":"6aa0c263-8349-11e7-89da-5254002ec671","apiVe...
			openshift.io/scc=privileged
Status:			Running
IP:			192.168.36.2
Controllers:		DaemonSet/glusterfs
Containers:
  glusterfs:
    Container ID:	docker://917da2ad277b5d51e5eca51600c7f04c7417867bdca3683cae84ed8391313a0d
    Image:		rhgs3/rhgs-server-rhel7:3.3.0-11
    Image ID:		docker-pullable://brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rhgs3/rhgs-server-rhel7@sha256:63f73ce4f13ee15429248c856c82b3487b5bb4df0a9638cacabbaaea6e90e266
    Port:		
    State:		Running
      Started:		Thu, 17 Aug 2017 08:41:47 -0400
    Ready:		True
    Restart Count:	0
    Limits:
      memory:	4Gi
    Requests:
      memory:		2Gi
    Liveness:		exec [/bin/bash -c systemctl status glusterd.service] delay=40s timeout=3s period=25s #success=1 #failure=15
    Readiness:		exec [/bin/bash -c systemctl status glusterd.service] delay=40s timeout=3s period=25s #success=1 #failure=15
    Environment:	<none>
    Mounts:
      /dev from glusterfs-dev (rw)
      /etc/glusterfs from glusterfs-etc (rw)
      /etc/ssl from glusterfs-ssl (ro)
      /run from glusterfs-run (rw)
      /run/lvm from glusterfs-lvm (rw)
      /sys/fs/cgroup from glusterfs-cgroup (ro)
      /var/lib/glusterd from glusterfs-config (rw)
      /var/lib/heketi from glusterfs-heketi (rw)
      /var/lib/misc/glusterfsd from glusterfs-misc (rw)
      /var/log/glusterfs from glusterfs-logs (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-07g37 (ro)
Conditions:
  Type		Status
  Initialized 	True 
  Ready 	True 
  PodScheduled 	True 
Volumes:
  glusterfs-heketi:
    Type:	HostPath (bare host directory volume)
    Path:	/var/lib/heketi
  glusterfs-run:
    Type:	EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:	
  glusterfs-lvm:
    Type:	HostPath (bare host directory volume)
    Path:	/run/lvm
  glusterfs-etc:
    Type:	HostPath (bare host directory volume)
    Path:	/etc/glusterfs
  glusterfs-logs:
    Type:	HostPath (bare host directory volume)
    Path:	/var/log/glusterfs
  glusterfs-config:
    Type:	HostPath (bare host directory volume)
    Path:	/var/lib/glusterd
  glusterfs-dev:
    Type:	HostPath (bare host directory volume)
    Path:	/dev
  glusterfs-misc:
    Type:	HostPath (bare host directory volume)
    Path:	/var/lib/misc/glusterfsd
  glusterfs-cgroup:
    Type:	HostPath (bare host directory volume)
    Path:	/sys/fs/cgroup
  glusterfs-ssl:
    Type:	HostPath (bare host directory volume)
    Path:	/etc/ssl
  default-token-07g37:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	default-token-07g37
    Optional:	false
QoS Class:	Burstable
Node-Selectors:	storagenode=glusterfs
Tolerations:	node.alpha.kubernetes.io/notReady=:Exists:NoExecute
		node.alpha.kubernetes.io/unreachable=:Exists:NoExecute
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath			Type		Reason		Message
  ---------	--------	-----	----			-------------			--------	------		-------
  30m		30m		1	kubelet, 192.168.36.2	spec.containers{glusterfs}	Normal		Pulled		Container image "rhgs3/rhgs-server-rhel7:3.3.0-11" already present on machine
  30m		30m		1	kubelet, 192.168.36.2	spec.containers{glusterfs}	Normal		Created		Created container
  30m		30m		1	kubelet, 192.168.36.2	spec.containers{glusterfs}	Normal		Started		Started container
  29m		29m		1	kubelet, 192.168.36.2	spec.containers{glusterfs}	Warning		Unhealthy	Liveness probe failed: ● glusterd.service - GlusterFS, a clustered file-system server
   Loaded: loaded (/usr/lib/systemd/system/glusterd.service; enabled; vendor preset: disabled)
   Active: inactive (dead)

  29m	29m	1	kubelet, 192.168.36.2	spec.containers{glusterfs}	Warning	Unhealthy	Readiness probe failed: ● glusterd.service - GlusterFS, a clustered file-system server
   Loaded: loaded (/usr/lib/systemd/system/glusterd.service; enabled; vendor preset: disabled)
   Active: inactive (dead)`

**Comment by MohamedAshiqrh** [comment]
@obnoxxx Updated the PR. Please review :)

**Comment by jarrpa** [comment]
@MohamedAshiqrh I just tried running with your patch, and I can't get the pods to deploy: my VMs don't have enough RAM to meet the 20GB request. :) I suggest removing the `request` value entirely, as I've been able to run and test the GlusterFS pods just fine with only 1GB of RAM per node.

**Comment by jarrpa** [comment]
@MohamedAshiqrh Also, why does your OCP output list the following values?
```
Limits:
memory: 4Gi
Requests:
memory: 2Gi
```

obnoxxx:
@MohamedAshiqrh, @jarrpa , may I also kindly remind you of the PR #195 that introduces an origin based openshift test env, and the would certainly need some amount of rebasing and reworking by now. :-)

I think as long as we ship openshift templates and an openshift mode, we might want to have such an env available ;-)

MohamedAshiqrh:
@jarrpa I tried with 4 Gi and 2 Gi as per my VM's capacity.

jarrpa:
@MohamedAshiqrh Okay, please remove the `request` value, then. :)

jarrpa:
@MohamedAshiqrh Specifying `limit` already inherently implies that Kube will try to prevent us from using more than that limit, so best I know we're not losing anything but removing `request`.

jarrpa:
@MohamedAshiqrh Reading @jeremyeder's suggestion in his comment, he indeed says to specify only a limit for memory but to only specify a request for CPU.

@jeremyeder Do we really need to target Burstable tier, or can we stay with Guaranteed? What's the difference? If we must be Burstable, will a CPU request + RAM limit get us there, and if so what's a good request level to have?

jeremyeder:
@sjenning can you please take a look?

sjenning:
@jarrpa @MohamedAshiqrh 

https://docs.openshift.com/container-platform/3.3/admin_guide/overcommit.html#qos-classes
https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

In order to be Guaranteed, a pod must specify a limit for all resources (cpu and memory).

If you specify a limit but not a request for a resource, the request is set equal to the limit.

A Guaranteed pod will be OOM killed if it tries to use more memory than its limit.  Its CPU is capped by CFS quota.

A Burstable pod is not capped in terms of CPU or Memory.  Its CPU is determined by cpu_shares (proportional to other pods), and memory is limited only by the node's Allocatable memory limit.   However, if the Burstable pod is way over its limit and eviction kicks in, it will be the first Burstable pod to  be evicted from the node after BestEffort Pods have been evicted.

tl;dr: Burstable is what you want here.  Specify requests for CPU and memory for the expected utilization.

jarrpa:
@sjenning ...so really, if we just specify a memory request of like `100Mi` then we're Burstable?

sjenning:
@jarrpa no you don't _need_ a CPU request, but you should provide one as a matter of principle so that the scheduler can see your pod as using some quantity of CPU.   I'd go with "100m" if your pod is idle most of the time.

jarrpa:
Superseded by #324

PR #221, opened 2017-03-30
Log command outputs in the abort function

obnoxxx:
Otherwise abort() will give no clues in the log.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/221)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -140,6 +140,7 @@ eval_output() {
     echo "return $?"
   )
 }
+
 abort() {
   ${CLI} delete svc heketi
   ${CLI} delete sa heketi-service-account
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -142,26 +142,26 @@ eval_output() {
 }
 
 abort() {
-  ${CLI} delete svc heketi
-  ${CLI} delete sa heketi-service-account
-  ${CLI} delete all,service,jobs,deployment,secret --selector="deploy-heketi"
-  ${CLI} delete svc/heketi-storage-endpoints
+  eval_output "${CLI} delete svc heketi 2>&1"
+  eval_output "${CLI} delete sa heketi-service-account 2>&1"
+  eval_output "${CLI} delete all,service,jobs,deployment,secret --selector=\"deploy-heketi\" 2>&1"
+  eval_output "${CLI} delete svc/heketi-storage-endpoints 2>&1"
   if [[ "${CLI}" == *oc\ * ]]; then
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -140,27 +140,28 @@ eval_output() {
...
-  ${CLI} delete secret heketi-db-backup
+  eval_output "${CLI} delete secret heketi-db-backup 2>&1"
   if [[ ${GLUSTER} -eq 1 ]]; then
     while read -r node; do
       debug "Removing label from '${node}' as a GlusterFS node."
-      ${CLI} label nodes "${node}" storagenode-
+      eval_output "${CLI} label nodes \"${node}\" storagenode- 2>&1"
     done <<< "$(echo -e "${NODES}")"
-    ${CLI} delete ds --selector="glusterfs"
+    eval_output "${CLI} delete ds --selector=\"glusterfs\" 2>&1"
```
jarrpa: I've found we don't need quotes around the selector. In this case, especially since its just a single word.

deploy/gk-deploy
```diff
@@ -140,27 +140,28 @@ eval_output() {
...
-    ${CLI} delete deployment heketi
+    eval_output "${CLI} delete deployment heketi 2>&1"
   fi
-  ${CLI} delete secret heketi-db-backup
+  eval_output "${CLI} delete secret heketi-db-backup 2>&1"
   if [[ ${GLUSTER} -eq 1 ]]; then
     while read -r node; do
       debug "Removing label from '${node}' as a GlusterFS node."
-      ${CLI} label nodes "${node}" storagenode-
+      eval_output "${CLI} label nodes \"${node}\" storagenode- 2>&1"
```
jarrpa: Did you verify this works? In particular I'm wondering if the quoting is preserved properly through eval.

obnoxxx: I did not. But there are other cases in the script which use this. e.g. Line 509:
```
eval_output "heketi-cli -s http://${heketi_service} --user admin --secret \"${ADMIN_KEY}\" topology load --json=${TOPOLOGY} 3>&1" | tee "${load_temp}"
```

But I guess i can also remove the quotes... in other cases ${node} is not quoted either in eval_output() calls...

deploy/gk-deploy
obnoxxx: correct. but they don't harm either (if they work...)
i can remove this one.

jarrpa:
OH, duh. Yeah, we can leave them then.

jarrpa:
LGTM

Issue #231, opened 2017-04-04
 don't delete a secret that we don't create


raghavendra-talur:
gk-deploy does not setup heketi with db backed up to secrets, hence it is not right that we delete the secrets for the same.

PR #232, opened 2017-04-04
don't delete a secret that we don't create

raghavendra-talur:
gk-deploy does not setup heketi with db backed up to secrets, hence it
is not right that we delete the secrets for the same.

Closes: #231

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/232)


jarrpa:
Fair enough. Looks like an artifact from when we went back to the dpeloy-heketi method.

LGTM

PR #242, opened 2017-04-07
The latest version of Openshift depreate -v 'value'

humblec:
with -p value in 'oc process' command. This patch enables
gk-deploy's backward compatibility based on oc version.

Signed-off-by: hchiramm <hchiramm@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/242)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -476,10 +476,14 @@ fi
 
 if [[ ${LOAD} -eq 0 ]]; then
   if [[ "${CLI}" == *oc\ * ]]; then
-    eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    oc_version = $(${CLI} version | grep "oc")
+    if [[ "${oc_version}" == *3.4.* ]]; then
+      eval_output "${CLI} process -v HEKETI_ADMIN_KEY=${ADMIN_KEY} -v HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    else
+      eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    fi
   else
     eval_output "${CLI} create -f ${TEMPLATES}/deploy-heketi-deployment.yaml 2>&1"
-  fi
 fi
 
 output -n "Waiting for deploy-heketi pod to start ... "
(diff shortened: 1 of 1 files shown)


humblec:
@raghavendra-talur  @obnoxxx @jarrpa  PTAL.

deploy/gk-deploy
```diff
@@ -476,10 +476,14 @@ fi
...
-    eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    oc_version = $(${CLI} version | grep "oc")
+    if [[ "${oc_version}" == *3.4.* ]]; then
+      eval_output "${CLI} process -v HEKETI_ADMIN_KEY=${ADMIN_KEY} -v HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    else
+      eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    fi
   else
     eval_output "${CLI} create -f ${TEMPLATES}/deploy-heketi-deployment.yaml 2>&1"
-  fi
```
obnoxxx: This removal of "fi" seems wrong: it creates unbalanced if .. fi brackets.

Locally this also fails  both the bash syntax and the shellcheck tests. not sure what happens in travis!? ...

obnoxxx: oh gosh, I found it: the travis environment does not have "realpath", which interestingly leads to run.sh exiting with exit code 0 without actually running any tests.... :-(

deploy/gk-deploy
```diff
@@ -476,10 +476,14 @@ fi
 
 if [[ ${LOAD} -eq 0 ]]; then
   if [[ "${CLI}" == *oc\ * ]]; then
-    eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
+    oc_version = $(${CLI} version | grep "oc")
+    if [[ "${oc_version}" == *3.4.* ]]; then
```
obnoxxx: One thought here: which versions do we support?
If we support origin (I think we do), oc reports v1.4.0 or v1.5.X, so this check would not apply.
We need to check for at least two flavors of oc version output  v1.X.Y and v3.X.Y.

It could be made more general like this:

```
VAL_SWITCH="-p"
oc_minor_version=$(oc version | grep oc | awk '{ print $2 }' | cut -d '.' -f2)
if [[ ${oc_minor_version} -lt 5 ]]; then
  VAL_SWITCH="-v"
fi
eval_output "${CLI} process ${VAL_SWITCH} HEKETI_ADMIN_KEY=${ADMIN_KEY} ${VAL_SWITCH} HEKETI_USER_KEY=${USER_KEY} deploy-heketi | ${CLI} create -f - 2>&1"
```

deploy/gk-deploy
```diff
@@ -551,7 +555,12 @@ check_pods "job-name=heketi-storage-copy-job" "Completed"
 eval_output "${CLI} delete all,service,jobs,deployment,secret --selector=\"deploy-heketi\" 2>&1"
 
 if [[ "${CLI}" == *oc\ * ]]; then
-  eval_output "${CLI} process -p HEKETI_ADMIN_KEY=${ADMIN_KEY} -p HEKETI_USER_KEY=${USER_KEY} heketi | ${CLI} create -f - 2>&1"
+    oc_version = $(${CLI} version | grep "oc")
```
obnoxxx: instead of duplicating the code, we should detect the version once at the top and set the VAL_SWITCH accordingly, and just use that whereever we do oc process... (i.e. use `${CLI} process ${VAL_SWITCH} KEY=VAL ...`)

deploy/gk-deploy
humblec: yeah, this was unintended. will correct it.

deploy/gk-deploy
humblec: Thanks. Not sure depending on minor version will work for earlier versions of origin or openshift.  also, I was first thinking to have this option as a command line switch and a variable set, but then we are introducing a new option  which is not known to existing users.  so dropped it. but above looks better, i will update patch set.

obnoxxx:
superseded by revised pr #244

deploy/gk-deploy
obnoxxx: Right, i *think* the current deployment scheme works for versions >= 3.4 / 1.4 only anyways, so it's probably not dangerous?

We might need to add additional checks when openshift deviates from the 3.X.Y / 1.X.Y pattern...

@humblec I did the work of updating the patch for you, see PR #244.

PR #245, opened 2017-04-07
gk-deploy: fix error in check_pods() when second argument is a string

obnoxxx:
The second argument of check_pods() can either be a timeout integer
value or a string to check for the status field. In the case of
a status string, it was mistakenly used as an override for the
timeout, leading to always error with timeout.

This patch lets check_pods inspect the second argument and only
use it as a timout override if it is numeric.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/245)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -187,7 +187,16 @@ assign() {
 
 check_pods() {
   local rc=1
-  local wait_limit=${2:-${WAIT}}
+  local wait_limit=${WAIT}
+  local number_re='^[0-9]+$'
+
+  # $2 can either be a status string or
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
This works, but seems like overkill. I would prefer not passing a string as $2 to check_pods()

obnoxxx:
@raghavendra-talur but that would require a custom treatment for the copy-job, right? Or how would you do that case?

raghavendra-talur:
I take back my comment. This is the easiest way to fix.

obnoxxx:
thanks @rtalur. merging..

PR #241, opened 2017-04-07
some gk-deploy fixes

obnoxxx:
Found a couple of smaller issues in gk-deploy.
The failure to error out when the copy-job fails rendered my setup invalid even though it was seemingly successful...

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/241)

PR #246, opened 2017-04-07
abort if pods don't come up

raghavendra-talur:
When storage-copy-job does not complete or when heketi pod does not
come up within timeout, we should not proceed with the deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/246)


obnoxxx:
this is already in #241

raghavendra-talur:
#241 fixes same issue

PR #247, opened 2017-04-07
wait for nodes to get to Ready state after joining

obnoxxx:
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/247)

### DIFF
--- a/vagrant/roles/nodes/tasks/main.yml
+++ b/vagrant/roles/nodes/tasks/main.yml
@@ -63,3 +63,11 @@
     remote_src: True
     owner: vagrant
     group: vagrant
+
+- name: wait for node to be ready
+  shell: test "$(kubectl get nodes {{ ansible_hostname }} --no-headers | awk '{ print $2 }')" = "Ready"
+  register: task_result
+  until: task_result.rc == 0
+  delay: 10
+  retries: 30
+  changed_when: false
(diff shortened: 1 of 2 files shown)


vagrant/roles/nodes/tasks/main.yml
```diff
@@ -33,3 +33,40 @@
...
+    remote_src: True
+    owner: vagrant
+    group: vagrant
+
+- name: wait for node to be ready
+  shell: test "$(kubectl get nodes {{ ansible_hostname }} --no-headers | awk '{ print $2 }')" = "Ready"
+  register: task_result
+  until: task_result.rc == 0
+  delay: 1
+  retries: 60
```
jarrpa: I'd recommend adding a `changed_when: False` here. Usually when doing a check or a wait it's common practice to not have a 'changed' state since it is effectively meaningless.

obnoxxx: cool! I did not know that, thanks! updated

not sure if delay 1 is too short. because it writes a message for each retry...

jarrpa: When i wait for pods (for example) I wait 10 seconds per retry. Let me know which you want to go with.

obnoxxx:
updated the delay to 5s  to be less chatty (it takes a couple of seconds for all to become Ready...)

obnoxxx:
updated to 10s wait as per @jarrpa  's recommendation

jarrpa:
LGTM!

PR #252, opened 2017-04-13
Tuning the Liveness and Readiness Probe options to support better scaling of gluster volume

MohamedAshiqrh:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/252)

### DIFF
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -50,20 +50,26 @@ spec:
           privileged: true
         readinessProbe:
           timeoutSeconds: 3
-          initialDelaySeconds: 60
+          initialDelaySeconds: 40
           exec:
             command:
             - "/bin/bash"
             - "-c"
             - systemctl status glusterd.service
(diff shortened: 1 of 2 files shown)


MohamedAshiqrh:
@obnoxxx @jarrpa @humblec @raghavendra-talur  @ramkrsna Please Review

MohamedAshiqrh:
Initial Time was 26 sec to be exact(Nothing to load in case of devicemapper as no custom fstab) for gluster pod to start.

So sticking to 40 delay seconds as we will have a heketidbstorage volume as default.

25 Period sec
15 Failure Threshold

This will give 

(25 x 15) + 40 = 415 seconds(~7 minutes) to fail in a failure case.

This will be good in case our scaling increasing for more volumes.

This change is required as for more scaled setup's the gluster pods fails to start.

It takes 

for 100 volumes, 100 seconds.
for 200 volumes, 160 seconds.
for 300 volumes, 220 seconds.

So the changes are based on this.

jarrpa:
LGTM!

PR #260, opened 2017-04-26
Add gk-installer Job and documentation

chaosaffe:
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/260)

### DIFF
--- a/installer-job/README.md
+++ b/installer-job/README.md
@@ -8,17 +8,26 @@ The repo includes an example ConfigMap and Job for deployment
 
 ### Step-by-step
 1. Create your [topology file](https://github.com/heketi/heketi/wiki/Setting-up-the-topology) for Heketi
+
 2. Create the Kubernetes ConfigMap with your topology file
+
   `kubectl create configmap gk-install-topology --from-file=<path to your topology file>`
+
 3. Create the Kubernetes Job to install gluster-kubernetes
+
   `kubectl create -f example/gk-install-job.yaml`
+
 4. Create the Kubernetes StorageClass referencing the GlusterFS/Heketi storage
+
   `kubectl create -f example/gk-storage-class.yaml`
 
 ### Building and Pushing the Docker image
 1. Build the image
+
   `docker build . -t <Docker Hub Username>/gk-installer`
+  
 2. Push the image
+
   `docker push <Docker Hub Username>/gk-installer`
 
-** Note: ** To use your image you will need to update `gk-install-job.yaml` to point to your container
+**Note:** To use your image you will need to update `example/gk-install-job.yaml` to point to your container


jarrpa:
The purpose of the Quickstart guide is to provide a narrow set of instructions to quickly experiment with gk-deploy. While I can see the benefit of having a job-based deployment strategy, it does not belong in the Quickstart guide (since part of the point is to get familiarity with the CLI tool).

That said, I'd love to see something like this make it into the repo. I'd have some requests:
 * Copy all the files (e.g. Dockerfile, ConfigMap) into this repo
 * Put all this in something like an `installer-job/` directory in the repo root
 * Provide a README with example instructions on how to use the Dockerfile to build the job image and how to create the various Kube resources. Make as few assumptions as possible on what the user knows how to do or has already done. :)
 * Instead of having a ConfigMap definition, provide instructions and/or maybe a script to create a ConfigMap from a topology file.

chaosaffe:
> While I can see the benefit of having a job-based deployment strategy, it does not belong in the Quickstart guide (since part of the point is to get familiarity with the CLI tool).

The intent behind the installer image is that the user does not need any familiarity with any CLI tools beyond `kubectl`. Ideally, the use of GluserFS/Heketi under the hood would be transparent to the user, assuming they deploy a StorageClass to Kubernetes (which I can include in the examples), instead allowing Kubernetes to provision any storage the user requires.

> Provide a README with example instructions on how to use the Dockerfile to build the job image 

Is there a possibility to instead use a provided Docker image, i.e. `gluster/gk-installer` instead of the user having to build their own image?

jarrpa:
> The intent behind the installer image is that the user does not need any familiarity with any CLI tools beyond kubectl. Ideally, the use of GluserFS/Heketi under the hood would be transparent to the user, assuming they deploy a StorageClass to Kubernetes (which I can include in the examples), instead allowing Kubernetes to provision any storage the user requires.

They can freaking deal. :) I'm all for having a StorageClass for pods to interact with, but getting to the point of having a StorageClass for any meaningful storage solution will involve work. My main issue with using an installer job is that troubleshooting when something goes wrong is much more difficult, as is re-running the installation with different parameters. Thus, I want the primary method for deployment to be using the shell script directly with a Job-based alternative as a convenience.

Additionally, on the pedantic notion of 'quick': Saying "write a config file and run this script against it" feels much more expedient than something like "write a config file, turn it into a ConfigMap, mount it as a volume in a job using our deployment image".

> Is there a possibility to instead use a provided Docker image, i.e. `gluster/gk-installer` instead of the user having to build their own image?

If we're going to provide an image that will ease this deployment, we are hosting the Dockerfile in this repo. We can provide a pre-built image on Docker Hub, no problem, but that image should come from this repo and we need to document how we got there so others can replicate it, however inane it may be. At that point, we should document the full workflow of going from Dockerfile in a repo to deployed storage solution.

installer-job/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Step-by-step
+1. Create your [topology file](https://github.com/heketi/heketi/wiki/Setting-up-the-topology) for Heketi
+
+2. Create the Kubernetes ConfigMap with your topology file
+
+  `kubectl create configmap gk-install-topology --from-file=<path to your topology file>`
+
+3. Create the Kubernetes Job to install gluster-kubernetes
+
```
jarrpa: Make an explicit mention/link to the example Job we provide, and that it can be used verbatim here.

installer-job/README.md
```diff
@@ -0,0 +1,33 @@
...
+  `kubectl create -f example/gk-install-job.yaml`
+
+4. Create the Kubernetes StorageClass referencing the GlusterFS/Heketi storage
+
```
jarrpa: Mention that you will need to discover and specify the heketi service URL.

installer-job/README.md
```diff
@@ -0,0 +1,33 @@
...
+  `kubectl create -f example/gk-storage-class.yaml`
+
+### Building and Pushing the Docker image
```
jarrpa: Put in something like `(optional)` at the end of the section header.

installer-job/README.md
```diff
@@ -0,0 +1,33 @@
+# gk-installer
+Installer for [gluster/gluster-kubernetes](https://github.com/gluster/gluster-kubernetes)
+
+## Usage
+The installer expects that you are running the Job in the Kubernetes cluster you intend to deploy gluster-kubernetes to. It also expects that you will mount your [topology file](https://github.com/heketi/heketi/wiki/Setting-up-the-topology) to /etc/gk-deploy/topology.json (i.e. via ConfigMap)
+
+The repo includes an example ConfigMap and Job for deployment
```
jarrpa: Mention that the ConfigMap we provide, as is, works with our example vagrant environment.

installer-job/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Step-by-step
+1. Create your [topology file](https://github.com/heketi/heketi/wiki/Setting-up-the-topology) for Heketi
+
+2. Create the Kubernetes ConfigMap with your topology file
+
+  `kubectl create configmap gk-install-topology --from-file=<path to your topology file>`
```
mrIncompetent: The filename of the topology must be `topology.json` as this is the filename the gk-installer pod is looking for

chaosaffe:
Stale. Closing

PR #264, opened 2017-05-12
should use namespace instead of context-name

FaKod:
context-name != current namespace for kubectl as well

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/264)


jarrpa:
LGTM!

PR #265, opened 2017-05-15
Using port forward

FaKod:
Changes:

1. I do not know, how you expect to access the deploy-heketi endpoint by this script. IMHO a port-forward to the heketi-pod is one solution for _stock_ Kubernetes (querying for free local port is missing...)
2. I had some remaining k8s objects, depending on if the NS is the current context or as cmd line parameter to the script
3. I think node label overwrite is valid here (had some errrors on script restart, if abort() wasn't called before)

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/265)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -581,7 +581,10 @@ if [[ ${?} != 0 ]]; then
   exit 1
 fi
 
+# this is for what is created by heketi-storage.json
 eval_output "${CLI} delete all,service,jobs,deployment,secret --selector=\"deploy-heketi\" 2>&1"
+# this ist for what is created before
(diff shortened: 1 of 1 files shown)


FaKod:
Hi @jarrpa, just out of interest, from where do you execute the script gk-deploy from?

jarrpa:
@FaKod Myself, I work primarily in VM environments. I run the gk-deploy script from a VM, either in or outside the kube cluster, that can access the cluster via kubectl.

FaKod:
ok, then I do not understand how you access your svc/heketi endpoints from outside the cluster (by gk-deploy).

Anyway, it's only for testing pods existence. 
So I close this PR now.

jarrpa:
@FaKod I deploy weave-net to facilitate my cluster networking. It allows the service to have an IP address accessible from outside the cluster.

PR #268, opened 2017-05-19
Design: Gluster-Block Provisioning End-To-End Design

obnoxxx:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/268)

### DIFF
--- a/docs/design/gluster-block-provisioning.md
+++ b/docs/design/gluster-block-provisioning.md
@@ -1,0 +1,383 @@
+# Gluster-Block Provisioning End-To-End Design
+
+This design is the result of the work of several
+people, most notably Humble Devassy Chirammal (@humblec),
+who started an ealier draft of this end-to-end design
+and gave many of the original ideas,
+but also Luis Pabon (@lpabon) who is largely
+responsible for an early cut at the heketi API,
+Steve Watt (@wattsteve),
+Vijay Bellur (@vbellur),
+Pranith Kumar Karampuri (@pranithk),
+and Prasanna Kumar Kalever (@pkalever).
+
+## Motivation
+
+While GlusterFS volumes can be used for RWO as well as RWX (and ROX) volumes,
+that support is not complete and problematic in some regards:
+
+1. RWO volumes should have protection from concurrent mounting.
+  (Locking to a host.) This is currently not enforced by the GlusterFS
+  mount plugin, hence the GlusterFS volumes may not meet the
+  expectations of the users. (This could be enforced in the GlusterFS
+  mount plugin, but there are more problems to solve.)
+2. The performance of the GlusterFS file volumes does not match the expectations
+  of many users who want to run dedicated workloads like databases and metrics
+  on the RWO volumes.
+  This is a fundamental problem in Gluster, because specifically metadata
+  operations are slow due to the distributed/shared nature.
+3. The scalability in the number of volumes that can be hosted is limited for
+  GlusterFS volumes. But because the RWO volumes can't be shared, specifically
+  for RWO we need a very good scalability. While the scalability in number
+  of volumes is improving for GlusterFS volumes, we need a different level
+  of scale.
+
+For these reasons, we need better, or rather *proper* RWO support
+that meets the above mentioned needs.
+
+## Goals
+
+The goal of the solution described in this design is to offer
+proper RWO support for Kubernetes PVs backed by Gluster.
+These improved RWO volumes should provide:
+
+* better separation in the sense that a volume can only be mounted
+  by one pod at a time.
+* better performance specifically for metadata-heavy workloads.
+* capability to scale to much larger numbers of volumes on one cluster.
+
+This new type of gluster volumes should be provisioned dynamically
+through storage classes.
+
+## Basic Idea and Analysis
+
+Use loopback files on GlusterFS volumes that are exported as block
+devices via iSCSI. Attach to these iSCSI devices and mount them on
+the Kubernetes nodes and bind-mount into the containers.
+
+This introduces an additional layer of indirection to
+file-system-access, so why would this be any faster?
+While data operations are very fast with Gluster, the meta-data
+operations are notoriously slow (GlusterFS being a distributed
+file system). But with the iSCSI loopback approach, all meta-data
+operations are translated into fast reads and writes.
+So this approach adds a small penalty to I/O operations, but
+gives a big speed-up to meta-data operations.
+
+Gluster's resource consumption is mainly by GlusterFS volume.
+Because this approach will use one gluster file volume to host
+many loopback files, it will allow us to support a much larger
+number of volumes with a given cluster and given hardware,
+easily scaling to thousands and ten thousands of volumes.
+
+A block device can not be used (R/W) by more than one
+entity at a time without risking data corruption. Hence
+for these RWO volumes, the single-mounter aspect needs to
+be enforced.
+
+## Components Involved
+
+The components involved in the overall flow from kubernetes through
+to gluster for this new functionality are:
+
+* A new external (out-of-tree) kubernetes provisioner ```glusterblock```,
+  to be exposed via a ```StorageClass```.
+* Gluster's RESTful service interface Heketi
+  with a new ```blockvolume``` set of subcommands to control the
+  creation of block volumes.
+* A new gluster tool ```gluster-block```, consisting of daemon component
+  and a command line tool. This tool is responsible for creating the
+  loopback files in a gluster volume and exporting them via iSCSI.
+* The kubernetes iSCSI mount plugin.
+
+## Overall Design and Flow
+
+The high-level flow for requesting and using a gluster-block based volume
+is as follows.
+
+* Administrator creates a ```StorageClass``` referring to the ```glusterblock```
+  provisioner.
+* User requests a new ```RWO``` volume with a ```PersistentVolumeClaim``` (PVC).
+* The ```glusterblock``` provisioner is invoked.
+* The provisioner calls out to ```heketi``` with the request to create a gluster
+  block volume with the requested characteristics like size, etc.
+* ```heketi``` looks for an appropriate gluster volume in one of the suitable
+  gluster clusters and calls out to ```gluster-block``` with the request to
+  create a block volume on the specified gluster file volume.
+* ```gluster-block``` creates a file on the gluster file volume and exports this
+  as an iSCSI target.
+* Upon success the resulting volume info (how to reach ...) is handed back
+  all the way to the provisioner.
+* The provisioner puts this volume information into a ```PersistentVolume```
+  (PV) and sets it up to use the iSCSI mount plugin for mounting.
+* The PV is bound to the original PVC.
+* The user references the PVC in an application description.
+* When the application pod is brought up on a node, the iSCSI mount plugin
+  takes care of initiating the block device, formatting (if required) and
+  mounting it.
+
+
+## Details About The ```glusterblock``` Provisioner
+
+The glusterblock provisioner is an external provisioner that is running in a
+container. It's code is located in the ```gluster/block``` subdirectory of
+<https://github.com/kubernetes-incubator/external-storage>.
+
+The provisioner is mainly a simple translation engine that turns PVC requests
+into requests for heketi via heketi's RESTful API, and wraps the resulting
+volume information into a PV.
+
+The provisioner is configured via a StorageClass which can look like this:
+
+```
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: glusterblock
+provisioner: gluster.org/glusterblock
+parameters:
+    resturl: "http://127.0.0.1:8081"
+    restuser: "admin"
+    secretnamespace: "default"
+    secretname: "heketi-secret"
+    hacount: "3"
+    clusterid: "630372ccdc720a92c681fb928f27b53f"
+```
+
+The available parameters are:
+
+* ```resturl```: how to reach heketi
+* ```restuser```, ```secretnamespace```, ```secretname```:
+  authentication information
+* ```hacount``` (optional): How many paths to the target server to configure.
+* ```clusterid``` (optional): List of one or more clusters to consider for
+  finding space for the requested volume.
+
+See  <https://github.com/kubernetes-incubator/external-storage/blob/master/gluster/block/README.md> for
+additional and up-to-date details.
+
+## Details About Heketi's New ```blockvolume``` Functionality
+
+For the purposes of gluster block volumes, the same heketi instance is used
+as for the regular glusterfs file volumes. Heketi has a new API though for
+treating blockvolume requests. Just as with the glusterfs file volume
+provisioning, the logic for finding suitable clusters and file volumes
+for hosting the loopback files is part of heketi.
+
+The API is a variation of the ```volume``` API and looks like this.
+
+The supported functions are:
+
+* ```BlockVolumeCreate```
+* ```BlockVolumeInfo```
+* ```BlockVolumeDelete```
+* ```BlockVolumeList```
+
+In the future, ```BlockVolumeExpand``` might get added.
+
+### Note About The State Of The Implementation
+
+As of 2017-06-23, a PR with the implementation is available
+at <https://github.com/heketi/heketi/pull/793>.
+
+
+### Details about the requests
+
+#### BlockVolumeCreateRequest
+
+The block volume create request takes the size and a name
+and can optionally take a list of clusters and an hacount.
+
+```
+type BlockVolumeCreateRequest struct {
+        Size       int       `json:"size"`
+        Clusters   []string  `json:"clusters,omitempty"`
+        Name       string    `json:"name"`
+        Hacount    int       `json:"hacount,omitempty"`
+        Auth       bool      `json:"auth,omitempty"
+}
+```
+
+#### BlockVolume
+
+This is the basic info about a block volume.
+
+```
+BlockVolume struct {
+        Hosts     []string `json:"hosts"`
+        Iqn       string   `json:"iqn"`
+        Lun       int      `json:"lun"`
+        Username  string   `json:"username"`
+        Password  string   `json:"password"`
+}
+
+```
+
+#### BlockVolumeInfo
+
+This is returned for the blockvolume info request and
+upon successful creation.
+
+```
+type BlockVolumeInfo struct {
+        Size       int       `json:"size"`
+        Clusters   []string  `json:"clusters,omitempty"`
+        Name       string    `json:"name"`
+        Hacount    int       `json:"hacount,omitempty"`
+        Id         string    `json:"id"`
+        Size       int       `json:"size"`
+        BlockVolume struct {
+                Hosts     []string `json:"hosts"`
+                Hacount   int `json:"hacount"`
+                Iqn       string `json:"iqn"`
+                Lun       int `json:"lun"`
+        } `json:"blockvolume"`
+}
+
+```
+
+
+#### BlockVolumeListResponse
+
+The block volume list request just gets a list
+of block volume IDs as response.
+
+
+```
+type BlockVolumeListResponse struct {
+        BlockVolumes []string `json:"blockvolumes"`
+}
+
+```
+
+### Details About Heketi's Internal Logic
+
+#### Block-hosting volumes
+
+The loopback files for block volumes need to be stored in
+gluster file volumes. Volumes used for gluster-block volumes
+should not be used for other purposes. For want of a better
+term, we call these volumes that can host block-volume
+loopback files **block-hosting file-volumes** or (for brevity)
+**block-hosting volumes** in this document.
+
+#### Labeling block-hosting volumes
+
+In order to satisfy a blockvolume create request, Heketi
+needs to find and appropriate block-hosting volume in
+the available clusters.  Hence heketi should internally
+flag these volumes with a label (`block`).
+
+#### Type of block-hosting volumes
+
+The block-hosting volumes should be regular
+3-way replica volumes (possibly distributed).
+One important aspect is that for performance
+reasons, sharding should be enabled on these volumes.
+
+#### Block-hosting volume creation automatism
+
+When heketi, upon receiving a blockvolume create request,
+does not find a block-hosting volume with sufficient
+space in any of the considered clusters, it would look for
+sufficient unused space in the considered clusters and create
+a new gluster file volume, or expand an existing volume
+labeled `block`.
+
+The sizes to be used for auto-creation of block-hosting
+volumes will be subject to certain parameters that can
+be configured and will have reasonable defaults:
+
+* `auto_create_block_hosting_volume`: Enable auto-creation of
+  block-hosting volumes?
+  Defaults to **false**.
+* `block_hosting_volume_size`: The size for a new block-hosting
+  volume to be created on a cluster will be the minimum of the value
+  of this setting and maximum size of a volume that could be created.
+  This size will also be used when expanding volumes: The amount
+  added to the existing volume will be the minimum of this value
+  and the maximum size that could be added.
+  Defaults to **1TB**.
+
+#### Internal heketi db format for block volumes
+
+Heketi stores information about the block volumes
+in it's internal DB. The information stored is
+
+* id: id of this block volume
+* name: name given to the volume
+* volume: the id of the block-hosting volume where the loopback file resides
+* hosts: the target ips for this volume
+
+#### Cluster selection
+
+By default, heketi would consider all available clusters
+when looking for space to create a new block-volume file.
+
+With the clusters request parameter, this search can be
+narrowed down to an explicit list of one or more clusters.
+With the help of the ```clusterid``` storage class option,
+this gives the kubernetes administrator a way to e.g. separate
+different storage qualities, or to reserve a cluster
+exclusively for block-volumes.
+
+
+### Details On Calling ```gluster-block```
+
+Heketi calls out to ```gluster-block``` the same way it
+currently calls out to the standard gluster cli for the
+normal volume create operations, i.e. it uses a kubexec
+mechanism to run the command on one of the gluster nodes.
+(In a non-kubernetes install, it uses ssh.)
+
+
+## Details About ```gluster-block```
+
+```gluster-block``` is the gluster-level tool to make creation
+and consumption of block volumes very easy. It consists of
+a server component ```gluster-blockd``` that runs on the gluster
+storage nodes and a command line client utility ```gluster-block```
+that talks to the ```gluster-blockd``` with local RPC mechanism
+and can be invoked on any of the gluster storage nodes.
+
+gluster-block takes care of creating loopback files on the
+specified gluster volume. These volumes are then exported
+as iSCSI targets with the help of the tcmu-runner mechanism,
+using the gluster backend with libgfapi. This has the big
+advantage that it is talking to the gluster volume directly
+in user space without the need of a glusterfs fuse mount,
+skipping the kernel/userspace context switches and the
+user-visible mount altogether.
+
+The supported operations are:
+
+* create
+* list
+* info
+* delete
+* modify
+
+Details about the gluster-block architecture can be found
+in the gluster-block git repository <https://github.com/gluster/gluster-block>
+and the original design notes
+<https://docs.google.com/document/d/1psjLlCxdllq1IcJa3FN3MFLcQfXYf0L5amVkZMXd-D8/edit?usp=sharing>.
+
+
+## Details About Using The In-Tree iSCSI Mount Plugin
+
+The kubernetes in-tree iSCSI mount plugin is used by the PVs created
+by the new glusterblock external provisioner. The tasks that are performed
+when a user brings up an application pod using a corresponding glusterblock
+PVC are:
+
+* create /dev/sdX on the host (iscsi login / initiator)
+* format the device if it is has not been formatted
+* mount the file system on the host
+* bind-mount the host mounted directory into the application pod
+
+Multi-pathing support has been added to the iSCSI plugin so that
+the gluster-block volumes are made highly availble with the ```hacount```
+feature.
+
+The iSCSI plugin has been changed in such a way that it does not allow
+more than one concurrent R/W mount of an iscsi block device.


obnoxxx:
This is the proposed end-to-end design for the gluster block volume support.
Pleas review, @humblec @lpabon @raghavendra-talur @MohamedAshiqrh @ramkrsna @jarrpa ,  ...

obnoxxx:
@wattsteve

obnoxxx:
This PR should only be merged after this design proposal is accepted.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,306 @@
+# Gluster-Block Provisioning End-To-End Design
+
+## Motivation
+
+While GlusterFS volumes can be used for RWO as well as RWX (and ROX) volumes,
+for some workloads, specifically metadata-heavy, the RWO support is not
+notsatisfactory with respect to performance. Furthermore, the scalability
```
vbellur: minor typo - spurious "not" needs to be removed.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,306 @@
...
+file system). But with the iSCSI loopback approach, all meta-data
+operations are translated into fast reads and writes.
+So this approach adds a small penalty to I/O operations, but
+gives a big speed-up to meta-data operations.
+
+Gluster's resource consumption is mainly by GlusterFS volume.
+Because this approach will use one gluster file volume to host
+many loopback files, it will allow us to support a much larger
+number of volumes with a given cluster and given hardware,
+easily scalin to thousands and ten thousands of volumes.
```
vbellur: s/scalin/scaling/

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,306 @@
...
+The supported operations are:
+
+* create
+* list
+* info
+* delete
+* modify
+
+Details about the gluster-block architecture can be found
+in the gluster-blocl git repository <https://github.com/gluster/gluster-block>
```
vbellur: s/bloc1/block/

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,306 @@
...
+With the help of the ```clusterid``` storage class option,
+this gives the kubernetes administrator a way to e.g. separate
+different storage qualities, or to reserve a cluster
+exclusively for block-volumes.
+
+#### Question
+
+Should Heketi create new file volumes for hosting block
+volume files (or in the future try to expand the currently
+used file volume, or is that solely up to the administrator?
```
vbellur: I feel that Heketi should create new file volumes if none of the existing volumes can satisfy the block request.

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: Yeah, I am inclined to that as well. But it poses a few questions:
* What sizes to allocate? (All of the available disk space? Half of it? 1 TB? Make that configurable? ...)
* Also auto-create the initial volume? (I'd think so.)

obnoxxx:
@vbellur thanks for the review! I fixed the typos.

Note that I added commits marked "SQ" on top of the original commit. This SQ should be read as "squash", and indicate that I intend to squash all these commits into the main commit before merging, once consensus is reached.

obnoxxx:
@vbellur, I have added details about an approach for auto-creation of block-hosting volumes. PTAL.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
+# Gluster-Block Provisioning End-To-End Design
+
+## Motivation
+
+While GlusterFS volumes can be used for RWO as well as RWX (and ROX) volumes,
+for some workloads, specifically metadata-heavy, the RWO support is not
+satisfactory with respect to performance. Furthermore, the scalability
+in number of volumes is currently not very good.
+
```
wattsteve: I think the scalability issue is well framed but I think the rest of the problem description/motivation is a little misleading. I think its important to clarify to folks inside and outside the community so they can understand what problem we're really solving here. In reality today, the gluster deployment that comes with gluster-kubernetes only supports ReadWriteMany volumes(RWX). RWO volumes are defined as only supporting a single concurrent writer and being locked to a host. While it is possible to use any RWX volume in place of a RWO volume, it does not provide the single writer or locking semantics that one expects when using a RWO volume. It is a bad idea to use a RWX volume in place of a RWO volume, because the Kubernetes user is unaware and they think they are getting the RWO semantics to protect their data from corruption and really they are not. Thus, this design PR is intended to add real RWO support to gluster-kubernetes.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+## Goals
+
+Offer better RWO support for Kubernetes PVs backed by Gluster.
```
wattsteve: remove "better". There is no RWO support today.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+The API is a variation of the ```volume``` API and looks like this.
+
+The supported functions are:
+
+* ```BlockVolumeCreate```
+* ```BlockVolumeInfo```
+* ```BlockVolumeDelete```
+* ```BlockVolumeList```
+
+In the future, ```BlockVolumeExpand``` might get added.
```
wattsteve: Yes, this makes sense, ExpandVolume or Resize is being targeted for Kubernetes 1.7, but the community has not landed on a design yet, so it makes sense to wait to add this operation

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+<https://github.com/obnoxxx/heketi/tree/block-wip>.
+
+
+### Details about the requests
+
+#### BlockVolumeCreateRequest
+
+The block volume create request takes the size and a name
+and can optionally take a list of clusters and an hacount.
+
```
wattsteve: Can you clarify what exactly gets provisioned here? Is this a raw block device with no filesystem on it? Are we relying on the step that creates the PV to specify to the ISCSI plugin what type of filesystem to put on it at mount time? The reason I bring this up is that we intend to support Raw Block Devices as PVs in Kube 1.8, so it would be good to plan for that in advance so that operators can specify whether they want Heketi to specify a FileSystem to be added in the PV or not.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+By default, heketi would consider all available clusters
+when looking for space to create a new block-volume file.
+
+With the clusters request parameter, this search can be
+narrowed down to an explicit list of one or more clusters.
+With the help of the ```clusterid``` storage class option,
+this gives the kubernetes administrator a way to e.g. separate
+different storage qualities, or to reserve a cluster
+exclusively for block-volumes.
+
```
wattsteve: While you can have multiple of these, it sounds like we're proposing a single trusted storage pool that one can provision volumes intended for RWX and RWO. I think this is going to be tricky for a capacity management (And thin provisioning), debugging, scaling and for a QoS perspective. I would like to suggest the idea of having separate Trusted Storage Pools (TSPs) for RWO, RWM and Object (which is inevitably coming later). So a TSP for RWO would have just one large single volume. A TSP for RWX would have multiple volumes. Much easier to isolate issues, debug and isolate resources to guarantee performance.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+* format the device if it is has not been formatted
+* mount the file system on the host
+* bind-mount the host mounted directory into the application pod
+
+Multi-pathing support has been added to the iSCSI plugin so that
+the gluster-block volumes are made highly availble with the ```hacount```
+feature.
+
+The iSCSI plugin has been changed in such a way that it does not allow
+more than one concurrent R/W mount of an iscsi block device.
```
wattsteve: Can we add sections on:
- how we'll be securing the RWO volumes?
- how we'll be scaling?

docs/design/gluster-block-provisioning.md
obnoxxx: Thanks for your review, @wattsteve!

I am happy to clarify the text a bit, but here are a few thoughts. Maybe it's me who has not really understood the problem we want to solve:

* RWO is defined as "the volume can be mounted as read-write by a single node": see https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes. (This sounds like a very very strange mode, because you can have as many writers as you want as long as they are on the same node. But even with that strange definition, I heard that this separation is not enforced by Kubernetes in any way.)
* Reading the definitions, a storage volume that can even be used for RWX purposes, can as well be exported as a RWO volume. Conceptually, there is nothing wrong about this. Hence we do provide perfectly valid RWO volumes with GlusterFS today.
* Also many users are using GlusfterFS backed RWO volumes today. So saying, we don't support RWO seems wrong to me.
* But for all I know, it is more the **expectations** that the users of RWO volumes have, specifically with respect to **performance**, that are not met by our GlusterFS-backed RWO volumes. This is because that capability of supporting multiple concurrent writers comes at a big cost for metadata-performance. So some users are happy with the performance, but others with more RWO-specific workloads are suffering from the performance. _This was the first and foremost reason for us to propose this design._
* I don't understand the comment about a RWX volume not providing the single-writer or locking semantics that you mention:
    * The file system of a RWO volume (i.e. with a single writer) would not need to care about locking as much as a RWX/shared volume where locking is crucially important. So what locking semantics does a RWO volume provide?
    * The file system itself usually can't do anything much about single-writer, right? This needs to be enforced by the kubernetes level if we want to do it, or what am I missing? (E.g. for the iSCSI mount plugin @humblec recently added the strict single-writer protection so that the lack of kube-wide enforcement mechanisms does not destroy our data...)

wattsteve: @obnoxxx Simply put, all RWO providers are expected to enforce locking semantics to a host. You are correct in that the AccessMode in k8s is purely descriptive. It tells the developer what semantics they are getting. Taking a RWX volume and describing it as RWO in the PV description will cause developers to submit claims against an RWO volume but instead get back a Shared FileSystem with no locking capabilities. So they think they are protected from another user in their same namespace using the claim in a different pod/container, but really they are not. It is not about performance, it is about locking to a host. Existing CNS volumes do not do this, thus they are not RWO.

obnoxxx: @wattsteve ok, now we are getting closer. I think we (I) may have a confusion of terminology. And I need to sort it out:

* For me, coming from storage, a distributed/shared file system like GlusterFS that is typically backing gluster, *does* have locking semantics that makes concurrent access from multiple writers possible. I.e. these distributed locking semantics are what makes RWX access safe.

* Different kinds of storage, like a local block device or local file system that might be a default candidate for RWO volumes do NOT have any locking semantics that prevent multiple mounts accessing it simultaneously from corrupting their data.

So it seems that with "locking semantics" you mean a mechanism that prevents the storage from being used more than one mount/writer or host. If that is the case, then (as I asked in my previous post) this protection mechanism would usually have to be enforced by the mount plugin of kubernetes, since the standard providers do explicitly LACK protection mechanisms. I think I get your point that due to the RWO flag, the users expect to not have to be careful when accessing the file system, since they assume they are alone.

(What by the way is the relevance of locking to a host instead of restricting to a single mount? I think this is a misconception that was introduced by the implementation detail that in most implementations PV mounts are done by a mount on the host and bind-mount into the container, hence all mounts of a PV on a host can be shared by a single host-mount via multiple bind-mounts. But that will still lead to data corruption due to unawareness, and hence we fixed the iSCSI plugin to lock to a single mount...)

So, coming back to the original point: This locking-to-a-mount could easily be achieved by the GlusterFS mount plugin, similarly to what is done in the iSCSI mount plugin. So, if it were only about those protection mechanisms, we would not be having this discussion.

Hence I stand by my point: The main point is performance.

What am I missing?

obnoxxx: Ok, we sorted it out. The challenges with RWO are three-fold:

1. RWO volumes should have protection from concurrent mounting. (Locking to a host.) This is currently not enforced for GlusterFS but could be achieved by modifying the GlusterFS mount plugin.
2. The performance of the GlusterFS file volumes does not match the expectations of many users who want to run dedicated workloads like databases and metrics on the RWO volumes.
    This is a fundamental problem in Gluster, because specifically metadata operations are slow due to the distributed/shared nature.
3. The scalability in the number of volumes that can be hosted is limited for GlusterFS volumes. But because the RWO volumes can't be shared, specifically for RWO we need a very good scalability.

I will clarify the text a bit in those lines.

docs/design/gluster-block-provisioning.md
obnoxxx: We thought about it. But we don't want to be that restrictive. The above comment explains how to achieve that separation by configuring different storage classes. It might be best practice to set it up that way. But the default should be the most simple solution where a single gluster cluster can host both regular GlusterFS PVs as well as block-hosting volumes with a lot of loopback devices.

screeley44: @obnoxxx - not sure I have a strong opinion yet, either way (i.e. single TSP or multiple TSP), I see the merits in both arguments.  I guess I'm leaning toward the flexibility argument, if an admin wants to create a single TSP that can handle shared, block and in the future object, then that is on them and is a good out of the box implementation for kicking the tires..But I also think realistically, that admins will also want to be able to set up multiple TSP as Steve mentioned so they can control capacity, service quality, etc.. So as long as there is no limitation on providing single or multiple TSP I think we are OK.  And basically we are defining a TSP as a clusterid in Heketi correct?  So a TSP is defined by a topology file that is loaded?  Just want to make sure I'm clear on the terminology.  The storageclasses can help control and isolate, based on quota limits, etc... which would be able to restrict namespaces from using certain storageclasses.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+upon successful creation.
+
+```
+type BlockVolumeInfo struct {
+        Size       int       `json:"size"`
+        Clusters   []string  `json:"clusters,omitempty"`
+        Name       string    `json:"name"`
+        Hacount    int       `json:"hacount,omitempty"`
+        Id         string    `json:"id"`
+        Size       int       `json:"size"`
```
MohamedAshiqrh: We can remove this.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,350 @@
...
+
+This is returned for the blockvolume info request and
+upon successful creation.
+
+```
+type BlockVolumeInfo struct {
+        Size       int       `json:"size"`
+        Clusters   []string  `json:"clusters,omitempty"`
+        Name       string    `json:"name"`
+        Hacount    int       `json:"hacount,omitempty"`
```
MohamedAshiqrh: There is two ha count here and 201.

obnoxxx:
I realized that not having attributions in the doc may have created the impression that I was the original author and the content and ideas are all mine. I fixed this by adding proper attribution to the beginning... :-)

I can also add signoff tags to the commit message for final squashing.

docs/design/gluster-block-provisioning.md
obnoxxx: @wattsteve, The kube raw block devices a good point! Are those raw block devices in kube intended for use in application pods? If so, we should definitely prepare.

We are planning to provision a raw block device here and let the iscsi provisioner format and mount it, as mentioned in the section on using the iscsi mount plugin below.

docs/design/gluster-block-provisioning.md
obnoxxx: I guess it has been moved out of kube 1.7, but planning for this future enhancement does not harm.

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done (replaced by "proper")

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+
+#### BlockVolumeInfo
+
+This is returned for the blockvolume info request and
+upon successful creation.
+
+```
+type BlockVolumeInfo struct {
+        Size       int       `json:"size"`
+        Clusters   []string  `json:"clusters,omitempty"`
```
raghavendra-talur: Why should we have a string of clusters here? It should be a singleton.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+
+#### BlockVolume
+
+This is the basic info about a block volume.
+
+```
+BlockVolume struct {
+        Hosts     []string `json:"hosts"`
+        Hacount   int `json:"hacount"`
+        Iqn       string `json:"iqn"`
```
raghavendra-talur: what does Iqn and Lun mean? Why are they not in Info?

lpabon:
Sorry it has taken me a while. I will be reviewing tomorrow.

@obnoxxx

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+## Motivation
+
+While GlusterFS volumes can be used for RWO as well as RWX (and ROX) volumes,
+that support is not complete and problematic in some regards:
+
+1. RWO volumes should have protection from concurrent mounting.
+  (Locking to a host.) This is currently not enforced by the GlusterFS
+  mount plugin, hence the GlusterFS volumes may not meet the
+  expectations of the users. (This could be enforced in the GlusterFS
+  mount plugin, but there are more problems to solve.)
```
humblec: @obnoxxx above seems really unclear to me. May be we need to put it in a way that `glusterFS` is a  shared FS or meant for many read/writers.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+be enforced.
+
+## Components Involved
+
+The components involved in the overall flow from kubernetes through
+to gluster for this new functionality are:
+
+* A new external (out-of-tree) kubernetes provisioner ```glusterblock```,
+  to be exposed via a ```StorageClass```.
+* Gluster's RestFUL service interface Heketi
```
humblec: RESTful

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Details About The ```glusterblock``` Provisioner
+
+The glusterblock provisioner is an external provisioner that is running in a
+container. It's code (currently WIP as of 2017-05-19) is located in the
+```gluster/block``` subdirectory of
+<https://github.com/kubernetes-incubator/external-storage>.
+
+The provisioner is mainly a simple translation engine that turns PVC requests
+into requests for heketi via heketi's RestFUL API, and wraps the resulting
```
humblec: RESTful

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+  by one pod at a time.
+* better performance specifically for metadata-heavy workloads.
+* capability to scale to much larger numbers of volumes on one cluster.
+
+This new type of gluster volumes should be provisioned dynamically
+through storage classes.
+
+## Basic Idea and Analysis
+
+Use loopback files on GlusterFS volumes that are exported as block
```
humblec: afaict, we cant say those are loopback files, rather its normal files in gluster volumes. @vbellur can confirm this.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+While data operations are very fast with Gluster, the meta-data
+operations are notoriously slow (GlusterFS being a distributed
+file system). But with the iSCSI loopback approach, all meta-data
+operations are translated into fast reads and writes.
+So this approach adds a small penalty to I/O operations, but
+gives a big speed-up to meta-data operations.
+
+Gluster's resource consumption is mainly by GlusterFS volume.
+Because this approach will use one gluster file volume to host
+many loopback files, it will allow us to support a much larger
```
humblec: same note as above on 'loopback' files.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+See  <https://github.com/kubernetes-incubator/external-storage/blob/master/gluster/block/README.md> for
+additional and up-to-date details.
+
+## Details About Heketi's New ```blockvolume``` Functionality
+
+For the purposes of gluster block volumes, the same heketi instance is used
+as for the regular glusterfs file volumes. Heketi has a new API though for
+treating blockvolume requests. Just as with the glusterfs file volume
+provisioning, the logic for finding suitable clusters and file volumes
+for hosting the loopback files is part of heketi.
```
humblec: same note on 'loopback' files.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+
+As of 2017-05-19, a WIP API and overall skeleton implementation is available in
+<https://github.com/obnoxxx/heketi/tree/block-wip>.
+
+
+### Details about the requests
+
+#### BlockVolumeCreateRequest
+
+The block volume create request takes the size and a name
```
humblec: IMO, `name` should not be 'mandatory' field of  BlockVolumeCreateRequest. It should be generated by Heketi as in file volume request if not filled.

docs/design/gluster-block-provisioning.md
humblec: @raghavendra-talur I dont understand what you meant here ? IQN and Lun are iscsi spec.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+        BlockVolumes []string `json:"blockvolumes"`
+}
+
+```
+
+### Details About Heketi's Internal Logic
+
+#### Block-hosting volumes
+
+The loopback files for block volumes need to be stored in
```
humblec: Same note on 'loopback' files.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+gluster file volumes. Volumes used for gluster-block volumes
+should not be used for other purposes. For want of a better
+term, we call these volumes that can host block-volume
+loopback files **block-hosting file-volumes** or (for brevity)
```
humblec: same note on loopback files.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+  Defaults to **1TB**.
+
+#### Internal heketi db format for block volumes
+
+Heketi stores information about the block volumes
+in it's internal DB. The information stored is
+
+* id: id of this block volume
+* name: name given to the volume
+* volume: the id of the block-hosting volume where the loopback file resides
```
humblec: same note on loopback files.

docs/design/gluster-block-provisioning.md
```diff
@@ -0,0 +1,387 @@
...
+## Details About ```gluster-block```
+
+```gluster-block``` is the gluster-level tool to make creation
+and consumption of block volumes very easy. It consists of
+a server component ```gluster-blockd``` that runs on the gluster
+storage nodes and a command line client utility ```gluster-block```
+that talks to the ```gluster-blockd``` with local RPC mechanism
+and can be invoked on any of the gluster storage nodes.
+
+gluster-block takes care of creating loopback files on the
```
humblec: same note on 'loopback' files.

docs/design/gluster-block-provisioning.md
obnoxxx: That is not the point: Sure glusterfs is *capable* of doing multiple concurrent writers. But it *could* certainly be used in way that there is only one writer. But the point is that this mode is currently not enforced by the glusterfs mount plugin for Kubernetes, even if the volume is flagged as an RWO volume. But for RWO volumes there is some expectation by the users that such restriction to single writer (or writers only from single node) is enforced by the mount plugin or file system.

docs/design/gluster-block-provisioning.md
obnoxxx: To my understanding, generally speaking, a loopback file is a file that is used as a block device (by some mechanisms, i.e. the loopback driver, or ...). Of course the files are just normal files. "loopback" refers to the way we are using them as block devices.

@humblec if you have a better wording in mind, please let me know.

docs/design/gluster-block-provisioning.md
obnoxxx: Thanks, will update this.

docs/design/gluster-block-provisioning.md
obnoxxx: you are right. this also is how we implement it ... Will change, thanks!

obnoxxx:
Thanks for the review, @humblec . I plan to send updates later today.

docs/design/gluster-block-provisioning.md
humblec: afaict, these are `sparse files`' in nature.  So may be we could word it something like that. @obnoxxx @vbellur thoughts?

obnoxxx: @humblec hmmm, well.
* "sparse" is a property of the files that describes how space is allocated inside the files. It is also an implementation detail: We could sparse or non-sparse files for our purposes.
* "loopback" in contrast refers to the purpose for which we want to use these files. And this was my intention.
The point I wanted to make by using a name for those files is that these are files (regular files in a gluster volume) that we use to provide "virtual block devices", which is afaik generally known as loop or loopback files/devices.

humblec:
@obnoxxx sure. Also,  when we first prepared the design doc, there was no CHAP authentication. It was added late to `gluster-block`.  I feel, we should include that information in the design doc.

lpabon:
@obnoxxx Thank you for putting this together.

I have been putting a lot of thought into this. At first, (almost a year ago) I thought that adjusting Heketi would be the right method, but now that I have deployed MySQL and CockroachDB on top of GlusterFS, I think a different model would make this much simpler.

Instead of adjusting Heketi and possibly de-stabilizing it, I would instead build on top of the stability we have with something new, call it glusterblock.

This new service would concentrate on block intelligence on top of Heketi. Let Heketi do what it is good at. Instead this services would provide the following APIs (as shown in this paper):

* Create, Info, Delete, List

## APIs

### Create

On Create, it would have the following algorithm:

1. Create a volume on GlusterFS using Heketi if one does not exist.
1. Using the K8S go-client API, create a new deployment of replica 1 for the iscsi bridge:
	* iscsi bridge would be started and it would check if the file exists or not, and create if needed.
	* Create iscsi setup
	* Wait for connection to serve iscsi

The Info,Delete, and List information can be kept as a Custom Resource Definitions (used to be called Third Party Resources) on Kubernetes. That way the application is stateless.

# Benefits

* It could be relesed asynchronously as a external provisioner because iSCSI is already part of the K8S plugin support
* Built on top of stable reliable systems
* Can allow for future use of losetup based block device instead of iSCSI once the CSI (Common Storage Interface) is complete from the CNCF (which will most likely replace Flexvols).

# Conclusion

I really want this feature! But I think editing Heketi is the wrong way around it. Instead, keep things encapsulated to their work and build on top of reliable systems. This will actually make it much faster to create.

jarrpa:
While conceptually I can see the benefits of such a model, it sounds like in practice this might result in a tight coupling between glusterblock and heketi anyway (in particular with the management of # of volumes, # of loopback files per volume, and volume options). Can you flesh it out a little more to see if it alleviates those concerns?

lpabon:
Hi @jarrpa , to quote Unix philosophy: *Make each program do one thing well. To do a new job, build afresh rather than complicate old programs by adding new "features".* 

Imagine if there was a need to also provide Object Storage. We could expand Heketi also provide the APIs to handle that, but it would make it much easier for something else to create the volume then place the object storage service on top. For example, I could see Minio or OpenStack Swift based object stores on top.

It is ok to couple glusterblock to Heketi, but the great thing about that is that if there are other models for block store on stop of glusterfs, then they can also be successful. Another great reason is encapsulation. Keeping all that knowledge onto a separate unit, will make software construction and testing much easier. This is actually the issue with creating large monolithic systems which are impossible to test.

Finally I think that it will make it much easier to deploy in Kubernetes, following the microservices model.

jarrpa:
And a lot of people really hate pure Unix-y tools. Any obsession with methodology in any direction is a problem, and there is such a thing as over-design. That said, it is very likely there are more concrete examples or intuitions that I'm simply not considering, so by all means take my comments with a grain of salt. :)

At the moment I see no practical benefits, just theoretical ones. If there are no other components/strategies that we would be interested in being compatible with, either now or in the near future, it still stands that this work could be part of heketi until such a time as there is an actual need for compatibility with another system. If it all is carefully designed anyway, the overhead of spawning it into its own project/component should be more than mitigated by the ease and speed of initial implementation. Testing and maintenance, as well, I would imagine would be easier with being in the same codebase at the outset.

obnoxxx:
@lpabon thanks for your comments. What you are proposing sounds very vague to me, though. What would that "glusterblock" service be?

Note that in our design we do have a kubernetes external provisioner, called "gluster/block". This does in fact use the iscsi mount plugin for mounting the resulting persistent volume. The provisioning calls out to heketi. The logic layer that creates the file volume (block-hosting) volume and creates the loopback files on the file volume is in this approach implemented as a new API in heketi. It could have been implemented as yet another daemon (in the unix philosophy), but many people were objecting to exactly that introduction of yet another daemon. They want less daemons. Also with tools like systemd, the old unix-philosophy is going away anyways. We seem to live in a post-unix world... ;-)

Finally, the most compelling argument you give, "This will actually make it much faster to create.", is vain because it is essentially already created:

* gluster-block is done: https://github.com/gluster/gluster-block
* the heketi work is essentially done: https://github.com/heketi/heketi/pull/793
* the external gluster/block provisioner exists https://github.com/kubernetes-incubator/external-storage (gluster/block) and the heketi plumbing is almost done.

Cheers - Michael

lpabon:
@obnoxxx I did not know about gluster-block. This is great! On Heketi #793, that is over 1.5k new lines. That will take time to review.

I think we should have a meeting to discuss this. I do not think we any changes to Heketi at all.  I'll send out an invite to the dev list.

lpabon:
Also, I'm not sure if you have seen this, just fyi: https://github.com/coreos/go-tcmu

obnoxxx:
@lpabon  if you did not know about gluster-block, then you have not carefully read this design proposal, and you have not followed gluster upstream either ;-)

obnoxxx:
@lpabon I don't understand yet, why you want to avoid changing heketi. I think this is precisely the right place for advanced logic of creating block-hosting volumes and placing block-files in them via gluster-block. This way it becomes a generic feature and kube-specific.

humblec:
@lpabon I still fail to understand how having a new block API in heketi could `destabilize heketi`? Gluster Block utility has 'more' logic than just setting up ISCSI server, also it has to attached or coupled with glusterd of the cluster.  Heketi already has logic of multi cluster awareness, talking to binaries via an different executor..etc.  So, why cant we make use of the existing logic in heketi when operating in different setups  ? IMO, that is better than splitting that logic to provisioner/controller and `gluster-block` and leaving heketi as is , Isnt it ? TPR, CDR..etc are in alpha/beta stage in upstream and CSI is in kind of design/poc phase.  Depending on something which is NOT yet stablized in upstream looks really risky rather I feel we should build  on currently available  stable systems.

lpabon:
Guys check out https://github.com/lpabon/poc-pvc-create may help in our discussions.

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: hmmm. for now keeping it like this. in the code we override with a generated name if the input is empty.

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: done

docs/design/gluster-block-provisioning.md
obnoxxx: no limitation, we provide max flexibility here, with most simple default

obnoxxx:
After (almost?) all of the requests have been addressed, I have just squashed all fix-up commits together, and I think this PR is now good to go.

jarrpa:
LGTM!

PR #270, opened 2017-06-08
update yum repo url and disable gpg check

raghavendra-talur:
The yum URL now redirects to a different place, let us change to that.
Also, the packages might not be signed always, removing gpg check there.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/270)


jarrpa:
LGTM

PR #272, opened 2017-06-11
[PROPOSAL] Codify release criteria and branch structures.

jarrpa:
This is an attempt to follow up on the conversation in #190 about component project versions. The main thing I see release branches for is to freeze the versions of the component projects, as well as make guarantees on stability of the code.

I'd like to hear at least from @humblec and @obnoxxx. The PR will be left open for several days after that to allow others to comment as well. At the end of either that period or when every dev has commented and signed off on it, if no major complaints have been raised it will be merged at that time. :)

### DIFF
--- a/docs/release-maintenance.md
+++ b/docs/release-maintenance.md
@@ -1,0 +1,87 @@
+# gluster-kubernetes Release and Maintenance Policies
+
+This document outlines the release and maintenance policies of the
+gluster-kubernetes project.
+
+## Maintained Branches
+
+The project will only support and actively maintain two branches, `master` and
+`stable`. Requests for support of older branches may be considered on a
+case-by-case basis, but users will be encouraged to use newer versions where
+possible.
+
+## Version Numbering
+
+This project follows the versioning guidelines outlined by the [Semantic
+Versioning specification, version 2.0.0](http://semver.org/spec/v2.0.0.html).
+In short, versions numbers will follow the structure `<MAJOR>.<MINOR>.<PATCH>`,
+with the following definitions:
+
+ * MAJOR version indicates a fundamental change to the structure of the
+   project, often due to innovations from significant changes in the component
+   projects. Major versions will typically not be compatible with older
+   versions of the component projects.
+ * MINOR version indicates a major feature, a broad set of changes, and/or new
+   releases of the component projects. Minor versions retain the following
+   compatibility guarantees:
+   1. Component projects from the last MAJOR release will still work with the
+      current code.
+   2. Deployments made with the current code will not conflict with other
+      deployments made since the last MAJOR release.
+ * PATCH version indicates backwards-compatible bug fixes, and guarantees that
+   the versions of the component projects has not changed.
+
+## Branch Definitions and Structure
+
+The `master` branch will always contain the latest development code. The
+project guarantees that this branch will be functional and tested but not
+bug-free. It will always track the latest versions of all component projects
+and makes no guarantee of backwards compatibility to older versions of those
+projects or itself.
+
+The `stable` branch will track the latest stable release of the code. When a
+new release is made, the `stable` branch will be moved to follow the new
+release branch.
+
+Each MAJOR and MINOR release will get its own branch, forked from `master`.
+Each release branch name will be of the form `<version>-latest`, e.g.
+`1.0-latest`. PATCH releases to those versions will be made in those branches,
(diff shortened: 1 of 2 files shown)


docs/release-maintenance.md
```diff
@@ -0,0 +1,87 @@
...
+and makes no guarantee of backwards compatibility to older versions of those
+projects or itself.
+
+The `stable` branch will track the latest stable release of the code. When a
+new release is made, the `stable` branch will be moved to follow the new
+release branch.
+
+Each MAJOR and MINOR release will get its own branch, forked from `master`.
+Each release branch name will be of the form `<version>-latest`, e.g.
+`1.0-latest`. PATCH releases to those versions will be made in those branches,
```
obnoxxx: I am confused. Further above you state that there will be only the `master` and `stable` branch, this is the first mention of a `MAJOR.MINOR-latest` branch. Requesting clarification.

docs/release-maintenance.md
```diff
@@ -0,0 +1,87 @@
...
+
+This project follows the versioning guidelines outlined by the [Semantic
+Versioning specification, version 2.0.0](http://semver.org/spec/v2.0.0.html).
+In short, versions numbers will follow the structure `<MAJOR>.<MINOR>.<PATCH>`,
+with the following definitions:
+
+ * MAJOR version indicates a fundamental change to the structure of the
+   project, often due to innovations from significant changes in the component
+   projects. Major versions will typically not be compatible with older
+   versions of the component projects.
```
obnoxxx: Just to understand, this means, e.g. when we introduced a new way of deploying gluster, e.g. with the DaemonSet, this would indicate a MAJOR version change (if we had done this versioning at the time) since previous versions of kubernetes supported in the earlier version of gluster-kubernetes, did not support DaemonSet -- correct?

docs/release-maintenance.md
obnoxxx: Apart from the one request for clarification, nice write-up!

humblec:
@jarpa, we dont have any compatibility matrix or mention about the depenedent upstream versions of  projects like `heketi`, `origin` and `kube. It would be better to draw a line between the upstream version dependency to the releases of this repo.

obnoxxx:
@humblec correct, we don't have any compat matrix or list. But it would be good to clearly state which (minimum) versions of component projects are required for a version of gluster-kubernetes. 

I don't understand what you mean by saying "It would be better to draw a line...".

humblec:
@obnoxxx what I meant was 'compatibility matrix' :)

jarrpa:
@humblec Creating such a compatibility matrix is a goal of this process. :)

docs/release-maintenance.md
jarrpa: Yes, that is the major example I had in mind.

docs/release-maintenance.md
jarrpa: Above I stated that `master` and `stable` are the only ones we would support and actively maintain. `stable` moves branches for every new release branch, but the old release branches stay in place (we just don't support them).

docs/release-maintenance.md
obnoxxx: good. makes sense!

docs/release-maintenance.md
obnoxxx: Hm, ok. So you are saying that the `stable` branch is basically just an alias for the newest `MAJOR.MINOR-latest` branch?

jarrpa: Yup. I can add a note to that effect under`Maintained Branches` if that would help.

obnoxxx: I think it would be helpful to state in the very beginning of the doc that we provide the master branch and a release branch for each stable MINOR.MAJOR version. And only the newest of those is further maintained under the alias of `stable`. I would find that less confusing. In the current doc, the `MINOR-MAJOR-latest` are mentioned too late imho.

jarrpa: Updated the text at the beginning:

The project will only support and actively maintain two branches, `master` and
the latest release branch. The latest release branch will always be reachable 
by two HEADS, either `<VERSION>-stable` or just `stable`.

jarrpa:
All right, it's been about a week. Taking it as no one having additional commentary. @obnoxxx, review?

jarrpa:
It's been over a month, so I'm merging this and we can change it later if someone has forgotten to speak up about it. :)

PR #273, opened 2017-06-11
Various vagrant updates

jarrpa:
A collection of minor fixes and major hacks I've come up with over time while using the vagrant environment for development, cleaned up for general public consumption. The latter commits can be moved to their own PRs if requested, but I figured I'd start simple. :)

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/273)

### DIFF
--- a/vagrant/README.md
+++ b/vagrant/README.md
@@ -14,3 +14,20 @@ Works in Vagrant/Ansible on Linux with libvirt/KVM or on Mac OS X on Virtualbox
 
 ## Versions
 Currently it uses Kubernetes v1.6.1
+
+## Features
+
+**CACHING:** This vagrant/ansible environment allows for caching of the yum
+cache and Docker images. This allows you to reuse those caches on subsequent
+provisioning of the VMs, and it is intended to help situations where one is
+developing in an environment where they would rather not have to redownload
+many megabytes repeatedly, e.g. hotel WiFi. :)  It stores both as `tgz` files
+in your `VAGRANT_HOME` directory, `~/.vagrant.d` by default. To enable this,
+either specify `VAGRANT_CACHE=1` on the command line or change the `CACHE`
+variable near the top of the Vagrantfile from `false` to `true`. **NOTE:** This
+will enable use of a Docker cache, but you must run `./docker_gen_cache.sh`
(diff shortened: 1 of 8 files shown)


vagrant/README.md
```diff
@@ -14,3 +14,20 @@ Works in Vagrant/Ansible on Linux with libvirt/KVM or on Mac OS X on Virtualbox
...
+**CACHING:** This vagrant/ansible environment allows for caching of the yum
+cache and Docker images. This allows you to reuse those caches on subsequent
+provisioning of the VMs, and it is intended to help situations where one is
+developing in an environment where they would rather not have to redownload
+many megabytes repeatedly, e.g. hotel WiFi. :)  It stores both as `tgz` files
+in your `VAGRANT_HOME` directory, `~/.vagrant.d` by default. To enable this,
+either specify `VAGRANT_CACHE=1` on the command line or change the `CACHE`
+variable near the top of the Vagrantfile from `false` to `true`. **NOTE:** This
+will enable use of a Docker cache, but you must run `./docker_gen_cache.sh`
+from the `vagrant/` directory to initialize the cache. This is because
```
obnoxxx: Why is this running of ./docker_gen_cache.sh not done automatically when the cache does not exist yet?

obnoxxx:
Nice work!
Just one question about the cache commit.

vagrant/README.md
jarrpa: It takes quite a bit of time to run, and it does not get faster on subsequent runs as the caching step is to export ALL container images into a tgz file. Running the caching every iteration would definitely be a considerable slowdown. I had considered doing the same for the yum cache as well, but I think it's not bad enough to warrant that.

obnoxxx: Ok. I thought you could conditionalize it with whometing like "when: !docker_cache_exists" or so...  ;-)

jarrpa: The other thing, too, is that I recommend pulling any other container images the user wants before generating the cache. I already pull GlusterFS and heketi automatically, but if they want like a containerized app cached as well they'd need to pull it to the master node first.

obnoxxx:
lgtm

PR #283, opened 2017-07-27
Add DeploymentConfig with replica for gluster-s3.

SaravanaStorageNetwork:
Add DeploymentConfig with replica for gluster-s3.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/283)

### DIFF
--- a/docs/examples/gluster-s3-storage-template/README.md
+++ b/docs/examples/gluster-s3-storage-template/README.md
@@ -214,7 +214,7 @@ oc rsh <glusters3pod>
 ```
 gswauth-prep -A http://<ipaddr>:8080/auth -K gswauthkey
 
-Where, ip address of the glusters3 pod obtained from 'oc get pods -o wide'
+Where, `ipaddr` is the IP address of the glusters3 pod obtained from 'oc get pods -o wide'
 ```
 
 #### To add user to account
--- a/docs/examples/gluster-s3-storage-template/README.md
+++ b/docs/examples/gluster-s3-storage-template/README.md
@@ -213,9 +213,9 @@ oc rsh <glusters3pod>
 #### This step prepares the gluster volume where gswauth will save its metadata
 ```
 gswauth-prep -A http://<ipaddr>:8080/auth -K gswauthkey
+```
 
 Where, `ipaddr` is the IP address of the glusters3 pod obtained from 'oc get pods -o wide'
-```
 
 #### To add user to account
 ```
--- a/docs/examples/gluster-s3-storage-template/README.md
+++ b/docs/examples/gluster-s3-storage-template/README.md
@@ -21,7 +21,7 @@ Available at
 ## Start glusters3 service using template
 
 ```
-oc new-app gluster-s3-template.yaml  --param=S3_ACCOUNT=testvolume  --param=S3_USER=adminuser --param=S3_PASSWORD=itsmine --param=VOLUME_CAPACITY=2Gi
+oc new-app gluster-s3-template.yaml  --param=GLUSTER_VOLUMES=testvolume  --param=GLUSTER_USER=adminuser --param=GLUSTER_PASSWORD=itsmine --param=VOLUME_CAPACITY=2Gi
 ```
 
 Note: adjust parameters according to your needs.
@@ -43,7 +43,7 @@ Available at:
 
 
 ```
-[root@master template]# oc new-app glusters3template.json  --param=S3_ACCOUNT=testvolume  --param=S3_USER=adminuser --param=S3_PASSWORD=itsmine --param=VOLUME_CAPACITY=2Gi
+[root@master template]# oc new-app glusters3template.json  --param=GLUSTER_VOLUMES=testvolume  --param=GLUSTER_USER=adminuser --param=GLUSTER_PASSWORD=itsmine --param=VOLUME_CAPACITY=2Gi      
 --> Deploying template "storage-project/glusters3template" for "glusters3template.json" to project storage-project
 
      glusters3template
@@ -52,9 +52,9 @@ Available at:
 
 
      * With parameters:
-        * S3 account=testvolume
-        * S3 user=adminuser
-        * S3 user authentication=itsmine
+        * Gluster volume=testvolume
+        * Gluster user=adminuser
+        * Gluster user authentication=itsmine
         * Volume capacity=2Gi
 
 --> Creating resources ...
--- a/docs/examples/gluster-s3-storage-template/README.md
+++ b/docs/examples/gluster-s3-storage-template/README.md
@@ -21,7 +21,7 @@ Available at
 ## Start glusters3 service using template
 
 ```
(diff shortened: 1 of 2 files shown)


SaravanaStorageNetwork:
@jarrpa @raghavendra-talur @humblec  Please review

SaravanaStorageNetwork:
@jarrpa 
About  GLUSTER_VOLUMES: 

For example to  access gluster-s3 object storage: 
s3curl.pl --debug --id "testvolume:adminuser" .. 

here,  testvolume is the object storage account we want to access. 

As mentioned, It is  "S3 storage account with underlying gluster volume" .

Gluster volume is mounted under  /mnt/gluster-object/${GLUSTER_VOLUMES}

When you are accessing "testvolume" object storage account, you are actually accessing GLUSTER_VOLUME mounted under /mnt/gluster-object/${GLUSTER_VOLUMES}

jarrpa:
Okay, so `GLUSTER_VOLUMES` is effectively the name of an account that will also be used for the name of a directory?

SaravanaStorageNetwork:
<<Okay, so GLUSTER_VOLUMES is effectively the name of an account that will also be used for the name of a directory? 
Yes, you are correct.

<< If so, this is not clear from the combination of the variable name and description. Would it make sense to change the variable name to something like S3_ACCOUNT and the description to S3 storage account that will be backed by a GlusterFS volume? 

Agreed. Naming is tough :) 
I think it makes sense to change other params too:
GLUSTER_USER as S3_USER
GLUSTER_PASSWORD as S3_PASSWORD
I will update all. 

<< Also, does the S3 container only ever create one GlusterFS volume for its use?
Yes, S3 container creates one GlusterFS volume.  This is just one example and it is possible to create more volumes if required.

jarrpa:
> Yes, S3 container creates one GlusterFS volume. This is just one example and it is possible to create more volumes if required.

Should we include a section on how to add more volumes and what that looks like?

SaravanaStorageNetwork:
> 
> Should we include a section on how to add more volumes and what that looks like?
> 

My understanding about template is, it helps you to launch an application (service).
This s3 template helps to launch s3 service.

If user is interested in launching same s3 service in another namespace, they can very well do it using the same template file. 

Do you want to add more s3 account( or configurable number of accounts) using  same template file ?  
I am not sure about that use case as of now.

But, with this existing template , you cannot add account once it is launched. 

I will check about this in future.

jarrpa:
The template as it stands only allows you to create the app in the namespace the template is in, thus you could need to recreate the template in a new namespace and use it there to deploy another app. Some suggestions:

* If you change the names of all the resources to `glusters3-${S3_ACCOUNT}-<resource type>` (e.g. `glusters3-${S3_ACCOUNT}-service`, you could use the same template to launch multiple GlusterS3 apps in the same namespace.
* Alternatively, is it theoretically possible to add more accounts to the same app by adding more `/mnt/gluster-object/<account>` mount points? Is there any additional env vars that could be added or modified to tell glusters3 to use these new mount points? If so you could tell the user to create a new PVC, then `oc edit` the DeploymentConfig to add the PVC as a new mount point.

SaravanaStorageNetwork:
@jarrpa  I will update the name changes as mentioned in future   (GLUSTER_VOLUMES as S3_ACCOUNT, GLUSTER_USER as S3_USER, GLUSTER_PASSWORD as S3_PASSWORD) as this involves changes docker rebuild.

> Some suggestions:
> 
> If you change the names of all the resources to glusters3-${S3_ACCOUNT}-<resource type> (e.g. glusters3-${S3_ACCOUNT}-service, you could use the same template to launch multiple GlusterS3 apps in the same namespace.

Nice idea..I will add this after the name changes done. 

> Alternatively, is it theoretically possible to add more accounts to the same app by adding more /mnt/gluster-object/<account> mount points? Is there any additional env vars that could be added or modified to tell glusters3 to use these new mount points? If so you could tell the user to create a new PVC, then oc edit the DeploymentConfig to add the PVC as a new mount point.

By executing these commands in the container, it is possible to add account into running container.

gswauth-prep -A http://<ipaddr>:8080/auth -K gswauthkey #this step Prepares the gluster volume where gswauth will save its metadata
gswauth-add-account -K gswauthkey <account> # This step is needed if account is not added into auth
gswauth-add-user -K gswauthkey -a <account> <user> <password>.

These commands needs to be executed in the running pod. 
I will dig further on this and see any better way to execute these commands in the running container.

For now, we will have this template file to launch one time S3 service with single account.

jarrpa:
> By executing these commands in the container, it is possible to add account into running container.
> 
> gswauth-prep -A http://:8080/auth -K gswauthkey #this step Prepares the gluster volume where gswauth will save its metadata
> gswauth-add-account -K gswauthkey # This step is needed if account is not added into auth
> gswauth-add-user -K gswauthkey -a .
> 
> These commands needs to be executed in the running pod.
> I will dig further on this and see any better way to execute these commands in the running container.
> 
> For now, we will have this template file to launch one time S3 service with single account.

I'm good with having the template launch with one account. Can you document the above steps in their own section (probably just before the testing commands) just so they're written down in the example?

SaravanaStorageNetwork:
@jarrpa  updated README

SaravanaStorageNetwork:
@jarrpa  Just to clarify -  By default authentication is enabled with one user. if required, another user  can be added. 
After adding the S3 user,  both the users can access the S3 account.

docs/examples/gluster-s3-storage-template/README.md
```diff
@@ -200,3 +203,21 @@ s3curl.pl --debug --id "testvolume:adminuser" --key "itsmine"  -- -k -v -s http:
...
+```
+oc rsh <glusters3pod>
+```
+
+#### This step prepares the gluster volume where gswauth will save its metadata
+```
+gswauth-prep -A http://<ipaddr>:8080/auth -K gswauthkey
+
+Where, ip address of the glusters3 pod obtained from 'oc get pods -o wide'
+```
```
jarrpa: Change the above three lines to:

```
gswauth-prep -A http://<ipaddr>:8080/auth -K gswauthkey
\```

Where, `ipaddr` is the IP address of the glusters3 pod obtained from 'oc get pods -o wide'
```

Minus the `\`.

SaravanaStorageNetwork: <<  Is there a reason you reverted your variable name change commit?
This needs changes in Dockerfile and few scripts in side docker container.

jarrpa: You didn't change the placement of the `` ``` `` line.

SaravanaStorageNetwork: @jarrpa  oops - updated now  - thanks for your patience :)

SaravanaStorageNetwork:
@jarrpa done storageclass related changes

jarrpa:
Woo! Thank YOU for your patience. :) LGTM!

PR #281, opened 2017-07-19
Use containerized heketi with non-containerized GlusterFS nodes

jarrpa:
This PR intends to allow gk-deploy to deploy a solution where heketi remains containerized and manages a set of non-containerized GlusterFS nodes. In its most basic form, this new functionality can be achieved by running:

```
gk-deploy -s <keyfile>
```

Where `<keyfile>` is a path to an SSH private key that can be used to get acces to either root or a sudoable account on the GlusterFS nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/281)

PR #285, opened 2017-08-04
[WIP] Deploy block and S3 containers by default

jarrpa:
Depends on: #281 


deploy/kube-templates/glusterblock-provisioner.yaml
```diff
@@ -0,0 +1,28 @@
...
+          glusterfs: glusterblock-provisioner-pod
+      spec:
+        serviceAccountName: glusterblock-provisioner
+        containers:
+        - name: glusterblock-provisioner
+          image: gluster/glusterblock-provisioner:latest
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: PROVISIONER_NAME
+            value: glusterblock-provisioner
```
jarrpa: I just went with what you had in the templates you provided. And per our earlier discussions, I see no need to make this configurable if it is very unlikely to change. If it ever changes, we can just update the scripts as I imagine this will not be a common occurrence. Does it matter what the value is, or is `gluster.org/glusterblock` correct? If so, why is it not so in the templates you have? :)

deploy/gk-deploy
```diff
@@ -94,6 +123,26 @@ Options:
...
+
+  --object-sc STORAGE_CLASS
+              Specify an optional, pre-existing StorageClass for the object
+              container to use to create GlusterFS volumes to host the objects.
+              Default is to create a new StorageClass.
+
+  --object-capacity CAPACITY
+              The total capacity of the GlusterFS volume which will host the
+              objects. Default is '${OBJ_CAPACITY}'.
+
```
jarrpa: If the user omits even one of the three auth parameters gluster-s3 can't be deployed and thus it is not deployed by default. The default capacity is 2Gi, as per your example.

deploy/kube-templates/glusterblock-provisioner.yaml
humblec: Yeah, I had it changed in the template to `gluster.org/glusterblock` , however failed to update you in the mail. so appreciated if you can keep this here as well.

deploy/kube-templates/glusterfs-daemonset.yaml
```diff
@@ -58,6 +60,9 @@ spec:
         securityContext:
           capabilities: {}
           privileged: true
+        env:
```
raghavendra-talur: need to rebase this patch on master and merge the env section with existing one.

jarrpa:
A small update: I am holding this PR in WIP status until the upstream Gluster containers and heketi are updated for gluster-block.

obnoxxx:
@jarrpa can this be merged now? :-)

PR #287, opened 2017-08-08
vagrant: improved method for caching Docker images

jarrpa:
Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/287)

PR #289, opened 2017-08-08
Introduce idempotent execution in gk-deploy

jarrpa:
This PR allows for gk-deploy to be rerun in the same environment and continue from where it determines to be where it left off. It also removes the default behavior to abort on error, allowing for faster troubleshooting of a failed environment and introducing the ability to manually fix it before rerunning the script to continue. This also removes the need for the `--load` parameter. Finally, I removed the requirement of having `heketi-cli` installed, opting to use the one in the heketi pods instead.

Note that this does not resolve the problem that we have where we can not recover from an inconsistent or corrupt heketi database. If this happens after the topology has been loaded, the user has no recourse but to run `-g --abort` and wipe the GlusterFS storage.

This PR also includes a few minor fixes I found along the way, mostly because I wanted to isolate the changes specific to the idempotent run introduction.

Depends on #281, #297

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/289)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -690,6 +690,9 @@ fi
 
 eval_output "${CLI} label --overwrite svc heketi-storage-endpoints glusterfs=heketi-storage-endpoints heketi=storage-endpoints"
 eval_output "${CLI} delete all,service,jobs,deployment,secret --selector=\"deploy-heketi\" 2>&1"
+if [[ "${CLI}" == *oc\ * ]]; then
+  eval_output "${CLI} delete dc,route,template --selector=\"deploy-heketi\" 2>&1"
+fi
 
 if [[ "${CLI}" == *oc\ * ]]; then
   eval_output "${CLI} process ${OC_PROCESS_VAL_SWITCH} HEKETI_EXECUTOR=${EXECUTOR} ${OC_PROCESS_VAL_SWITCH} HEKETI_ADMIN_KEY=${ADMIN_KEY} ${OC_PROCESS_VAL_SWITCH} HEKETI_USER_KEY=${USER_KEY} heketi | ${CLI} create -f - 2>&1"
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -579,7 +579,7 @@ if [[ ${LOAD} -eq 0 ]]; then
   fi
 fi
 
-if [[ $GLUSTER -eq 1 ]] && [[ ${LOAD} -eq 0 ]]; then
+if [[ ${GLUSTER} -eq 1 ]] && [[ ${LOAD} -eq 0 ]]; then
   if [[ -z ${DAEMONSET_LABEL} ]]; then
     DAEMONSET_LABEL=glusterfs
   fi
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -99,14 +99,13 @@ Options:
               deployment and re-run topology load command.
 
   --admin-key ADMIN_KEY
-              Secret for heketi administrator user. Heketi admin has access to all
-              the APIs. admin key: string, Shared secret.
-              Eg: MySecret
+              Secret string for heketi admin user. heketi admin has access to
+              all APIs and commands. Default is to use no secret.
 
   --user-key USER_KEY
-              Secret for heketi general user. Heketi user has access to only Volume
-              APIs. user key: string, Shared secret.
-              Eg: MyVolumeSecret
+              Secret string for general heketi users. heketi users have access
+              to only Volume APIs. Used in dynamic provisioning. Default is to
+              use no secret.
 
   --daemonset-label DAEMONSET_LABEL
               Controls the value of the label set on nodes which will host pods
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -28,7 +28,9 @@ WAIT=300
 ABORT=0
 NODES=""
 SKIP_PREREQ=0
-LOAD=0
+EXISTS_GLUSTERFS=0
+EXISTS_DEPLOY_HEKETI=0
+EXISTS_HEKETI=0
 EXECUTOR="kubernetes"
 SSH_KEYFILE="/dev/null"
 SSH_USER="root"
@@ -40,7 +42,7 @@ DAEMONSET_LABEL=""
 
 usage() {
   echo -e "USAGE: ${PROG} [-ghvy] [-c CLI] [-t <TEMPLATES>] [-n NAMESPACE] [-w <SECONDS>]
-       [--load] [-s <KEYFILE>] [--ssh-user <USER>] [--ssh-port <PORT>]
+       [-s <KEYFILE>] [--ssh-user <USER>] [--ssh-port <PORT>]
        [--admin-key <ADMIN_KEY>] [--user-key <USER_KEY>] [-l <LOG_FILE>]
        [--daemonset-label <DAEMONSET_LABEL> ]  [<TOPOLOGY>]\n"
 }
@@ -92,11 +94,6 @@ Options:
 
   -w SECONDS, --wait SECONDS
               Wait SECONDS seconds for pods to become ready. Default is '${WAIT}'.
-
-  --load      Resumes from the Topology load. Used in case of topology load
-              failure on adding nodes or devices to heketi. Fix the issue on
-              device or node, then restart with --load to skip gluster
-              deployment and re-run topology load command.
 
   --admin-key ADMIN_KEY
               Secret string for heketi admin user. heketi admin has access to
@@ -378,10 +375,6 @@ while [[ $# -ge 1 ]]; do
         h*|-help)
         help_exit
         ;;
-        -load)
-        LOAD=1
-        keypos=$keylen
-        ;;
         v*|-verbose)
         VERBOSE=1
         if [[ "$key" == "--verbose" ]]; then keypos=$keylen; fi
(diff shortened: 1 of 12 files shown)


obnoxxx:
If I see it correctly, all patches except for the last two are general fix patches, and not directly related to making the code idempotent, right? (the first one was just merged separately)
Couldn't the fix patches that do not implement idempotency be (also) pushed as separate prs?
A few comments on those preliminary patches:

* clean up resource labels: what's the motivation for this? the commit msg does not explain that...
* only exit on failure, don't abort: also here - what's the reasoning? (into commit msg)
* delete all OpenShift deploy-heketi resources: lgtm, could be merged immediately
* add missing curly braces: lgtm, could be merged right away
* clarify ADMIN...: lgtm, could be merged right away
* remove need for installing heketi-cli: interesting! certainly an improvement of its own. need to have a deeper look

jarrpa:
I kept them all in one PR because I didn't want to create a long dependency chain across them. Having a single PR depend on something like 6 other ones is annoying. :)

Updated the commit messages.

jarrpa:
More specifically, the changes for idempotency are in some way layered on top of all the other changes, such that they will not apply to the git tree without these other changes.

deploy/gk-deploy
```diff
@@ -515,14 +508,37 @@ if [[ "${EXECUTOR}" == "ssh" ]]; then
...
+  EXISTS_HEKETI=1
+  output "found."
+else
+  output "not found."
+fi
+
+if [[ ${EXISTS_HEKETI} -eq 0 ]]; then
+  output -n "Creating initial resources ... "
   if [[ "${CLI}" == *oc\ * ]]; then
     eval_output "${CLI} create -f ${TEMPLATES}/deploy-heketi-template.yaml 2>&1"
```
obnoxxx: Anyways the appearance of deploy-heketi here and further down confuses me...

deploy/gk-deploy
```diff
@@ -561,9 +578,10 @@ if [[ ${GLUSTER} -eq 1 ]] && [[ ${LOAD} -eq 0 ]]; then
     exit 1
   fi
   output "OK"
+  EXISTS_GLUSTERFS=1
 fi
 
-if [[ ${LOAD} -eq 0 ]]; then
+if [[ ${EXISTS_DEPLOY_HEKETI} -eq 0 ]] && [[ ${EXISTS_HEKETI} -eq 0 ]]; then
   sed -e "s/\${HEKETI_EXECUTOR}/${EXECUTOR}/" -e "s/\${SSH_PORT}/${SSH_PORT}/" -e "s/\${SSH_USER}/${SSH_USER}/" -e "s/\${SSH_SUDO}/${SSH_SUDO}/" heketi.json.template > heketi.json
```
obnoxxx: Here is more deploy-heketi stuff. What's the connection?

deploy/gk-deploy
jarrpa: In this case, no. We want to guarantee we'll always run this block unless we detect that the deployment is fully complete. The deployment is only complete if the final heketi pod exists.

deploy/gk-deploy
jarrpa: I'm not sure I see where you're confused. This block deploys deploy-heketi, thus if neither deploy-heketi nor heketi are running we should deploy it.

deploy/gk-deploy
obnoxxx: hmm, why would we want to always create the deploy-heketi thing when heketi does not exist? maybe we already created deploy-heketi?

deploy/gk-deploy
obnoxxx: but isn't deploy-heketi already deployed above? (cli create -f deploy-heketi-template ...)

deploy/gk-deploy
jarrpa: Two reasons.

1. It is possible that we have already successfully deployed deploy-heketi, completed the setup-openshift-heketi-storage command, deleted the deploy-heketi resources, and then failed or exited the script somehow. Since everything could be already setup for the final heketi pod, we should make sure these are always present.

2. This is looking ahead to the possibility of allowing the option to use a lube secret foe the heketi db, and thus would not need deploy-heketi

deploy/gk-deploy
jarrpa: That just creates the template, you later have to "process" the template to generate the final yaml that defines the actual resources.

obnoxxx: Ah, that makes sense. And it only proves that I may know something about kubernetes, but I know at least little about the differences in openshift... ;-)

deploy/gk-deploy
```diff
@@ -572,107 +590,112 @@ if [[ ${LOAD} -eq 0 ]]; then
...
-fi
+if [[ ${EXISTS_HEKETI} -eq 0 ]]; then
+  if [[ "${CLI}" == *oc\ * ]]; then
+    eval_output "${CLI} process ${OC_PROCESS_VAL_SWITCH} HEKETI_EXECUTOR=${EXECUTOR} ${OC_PROCESS_VAL_SWITCH} HEKETI_ADMIN_KEY=${ADMIN_KEY} ${OC_PROCESS_VAL_SWITCH} HEKETI_USER_KEY=${USER_KEY} heketi | ${CLI} create -f - 2>&1"
+  else
+    eval_output "sed -e 's/\\\${HEKETI_EXECUTOR}/${EXECUTOR}/' -e 's/\\\${HEKETI_ADMIN_KEY}/${ADMIN_KEY}/' -e 's/\\\${HEKETI_USER_KEY}/${USER_KEY}/' ${TEMPLATES}/heketi-deployment.yaml | ${CLI} create -f - 2>&1"
+  fi
 
-output -n "Waiting for heketi pod to start ... "
-check_pods "glusterfs=heketi-pod"
```
obnoxxx: This change check_pods "glusterfs=heketi-pod" --> check_pods "heketi=pod" looks like it should belong to the label cleanup patch...

deploy/gk-deploy
```diff
@@ -511,14 +504,37 @@ if [[ "${EXECUTOR}" == "ssh" ]]; then
   done <<< "$(echo -e "${NODES}")"
 fi
 
-if [[ ${LOAD} -eq 0 ]]; then
-  output -n "Checking that heketi pod is not running ... "
-  check_pods "glusterfs=heketi-pod" 2
```
obnoxxx: label cleanup patch?

deploy/gk-deploy
```diff
@@ -572,107 +590,112 @@ if [[ ${LOAD} -eq 0 ]]; then
   else
     eval_output "sed -e 's/\\\${HEKETI_EXECUTOR}/${EXECUTOR}/' -e 's/\\\${HEKETI_ADMIN_KEY}/${ADMIN_KEY}/' -e 's/\\\${HEKETI_USER_KEY}/${USER_KEY}/' ${TEMPLATES}/deploy-heketi-deployment.yaml | ${CLI} create -f - 2>&1"
   fi
-fi
 
-output -n "Waiting for deploy-heketi pod to start ... "
-check_pods "glusterfs=heketi-pod"
```
obnoxxx: label cleanup patch?

jarrpa:
@obnoxxx Good catch on the label cleanups. Fixed.

obnoxxx:
@jarrpa , this needs rebase / conflict resolution after merging #287 .

obnoxxx:
... and after merging #303 as well ;-)

obnoxxx:
@jarrpa i rebased it myself: https://github.com/obnoxxx/gluster-kubernetes/tree/pr-289

Ready to merge this, but i'll let you update this pr.

jarrpa:
Rebased and pushed

obnoxxx:
@jarrpa , you destroyed the two patches adding skip-preflight checks.
Here's the diff between my rebased branch and your branch:

```
$ git di HEAD..pr-289-2
 vagrant/roles/master/tasks/main.yml | 2 +-
 vagrant/roles/nodes/tasks/main.yml  | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)

diff --git a/vagrant/roles/master/tasks/main.yml b/vagrant/roles/master/tasks/main.yml
index b3c60d8..f7afd6c 100644
--- a/vagrant/roles/master/tasks/main.yml
+++ b/vagrant/roles/master/tasks/main.yml
@@ -5,7 +5,7 @@
 
 - block:
   - name: kubeadm init
-    command: kubeadm init --skip-preflight-checks --token={{ kubernetes_token }} --kubernetes-version=v{{ kubernetes_version }}  --apiserver-advertise-address={{ ansible_eth1.ipv4.address }}
+    command: kubeadm init --token={{ kubernetes_token }} --kubernetes-version=v{{ kubernetes_version }}  --apiserver-advertise-address={{ ansible_eth1.ipv4.address }}
 
   - name: create root kube dir
     file:
diff --git a/vagrant/roles/nodes/tasks/main.yml b/vagrant/roles/nodes/tasks/main.yml
index ad6a133..8617c57 100644
--- a/vagrant/roles/nodes/tasks/main.yml
+++ b/vagrant/roles/nodes/tasks/main.yml
@@ -40,7 +40,7 @@
 
 - block:
   - name: kubeadm join with master
-    command: kubeadm join --skip-preflight-checks --token={{ kubernetes_token }} {{ hostvars['master'].ansible_eth1.ipv4.address }}:6443
+    command: kubeadm join --token={{ kubernetes_token }} {{ hostvars['master'].ansible_eth1.ipv4.address }}:6443
 
   - name: create root kube dir on node
     file:

```

obnoxxx:
LGTM, thanks!

PR #299, opened 2017-08-16
gk-deploy: Deploy gluster-s3 container

jarrpa:
Depends on: #289, #305

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/299)

PR #302, opened 2017-08-17
Major enhancements to testing framework

jarrpa:
This PR seeks to introduce enhancements to the testing framework that achieve the following:

 * Improves coverage of the simple tests
 * Improve output of complex tests (now with colors!)

And a few other minor fixes and cleanups along the way.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/302)

### DIFF
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -36,8 +36,23 @@ fail() {
 }
 
 create_vagrant() {
-	cd ${VAGRANT_DIR}
-	./up.sh
+	cd "${VAGRANT_DIR}" || exit 1
+
+	local vstatus
+	local run=0
+
+	vstatus=$(vagrant status | grep "master\|node")
+	for m in ${vstatus}; do
+		if [[ "$(echo "${mstatus}" | grep "running")" == "" ]]; then
+			run=1
+		fi
+	done
+
+	if [[ ${run} -eq 1 ]]; then
+		./up.sh
+	fi
+
+        ssh_config
 }
 
 start_vagrant() {
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -83,22 +83,22 @@ create_vagrant() {
 }
 
 start_vagrant() {
-	cd ${VAGRANT_DIR}
+	cd "${VAGRANT_DIR}" || exit 1
 	vagrant up --no-provision || end_test -e "Error starting vagrant environment"
 }
 
 stop_vagrant() {
-	cd ${VAGRANT_DIR}
+	cd "${VAGRANT_DIR}" || exit 1
 	vagrant halt || end_test -e "Error halting vagrant environment"
 }
 
 destroy_vagrant() {
-	cd ${VAGRANT_DIR}
+	cd "${VAGRANT_DIR}" || exit 1
 	vagrant destroy || end_test -e "Error destroying vagrant environment"
 }
 
 ssh_config() {
-	cd ${VAGRANT_DIR}
+	cd "${VAGRANT_DIR}" || exit 1
 	vagrant ssh-config > "${SSH_CONFIG}" || end_test -e "Error creating ssh-config"
 }
 
@@ -115,9 +115,9 @@ rollback_vagrant() {
 }
 
 copy_deploy() {
-	cd ${VAGRANT_DIR}
 	local node=${1:-master}
 
+	cd "${VAGRANT_DIR}" || exit 1
 	scp -qr -F "${SSH_CONFIG}" "${DEPLOY_DIR}" "${node}:" || end_test -e "SCP deploy to ${node} failed"
 	scp -qr -F "${SSH_CONFIG}" "${TOPOLOGY_FILE}" "${node}:deploy/topology.json" || end_test -e "SCP topology to ${node} failed"
 }
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 # honor variables set from the caller:
-: ${TEST_DIR:="$(realpath $(dirname $0))"}
-: ${BASE_DIR:="${TEST_DIR}/../.."}
-: ${VAGRANT_DIR:="${BASE_DIR}/vagrant"}
-: ${DEPLOY_DIR:="${BASE_DIR}/deploy"}
-: ${TOPOLOGY_FILE:="${DEPLOY_DIR}/topology.json.sample"}
-: ${TESTNAME:="$(basename $0)"}
+: ${TEST_DIR="$(realpath $(dirname ${0}))"}
+: ${BASE_DIR="${TEST_DIR}/../.."}
+: ${VAGRANT_DIR="${BASE_DIR}/vagrant"}
+: ${DEPLOY_DIR="${BASE_DIR}/deploy"}
+: ${TOPOLOGY_FILE="${DEPLOY_DIR}/topology.json.sample"}
+: ${TESTNAME="$(basename ${0})"}
 
 SSH_CONFIG=${VAGRANT_DIR}/ssh-config
 
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -83,10 +83,14 @@ copy_deploy() {
 }
 
 pull_docker_image() {
-	local image=$1
-	cd ${VAGRANT_DIR}
-	for NODE in node0 node1 node2 ; do
-		ssh -F ${SSH_CONFIG} ${NODE} "sudo docker pull ${image}"
+	cd "${VAGRANT_DIR}" || exit 1
+
+	local image=${1}
+	local vstatus
+
+	vstatus=$(vagrant status | grep "node" | awk '{print $1}')
+	for NODE in ${vstatus}; do
+		ssh -q -F "${SSH_CONFIG}" "${NODE}" "sudo docker pull ${image}"
 	done
 }
 
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -102,6 +102,18 @@ ssh_config() {
 	vagrant ssh-config > "${SSH_CONFIG}" || end_test -e "Error creating ssh-config"
 }
 
+rollback_vagrant() {
+	cd "${VAGRANT_DIR}" || exit 1
+	(
+        ./rollback.sh
+	if [[ ${?} -ne 0 ]]; then
+		destroy_vagrant
+		create_vagrant
+		ssh_config
+	fi
+        ) || end_test -e "Error rolling back vagrant environment"
+}
+
 copy_deploy() {
 	cd ${VAGRANT_DIR}
 	local node=${1:-master}
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -1,38 +1,65 @@
 #!/bin/bash
 
 # honor variables set from the caller:
-: ${TEST_DIR="$(realpath $(dirname ${0}))"}
-: ${BASE_DIR="${TEST_DIR}/../.."}
-: ${VAGRANT_DIR="${BASE_DIR}/vagrant"}
-: ${DEPLOY_DIR="${BASE_DIR}/deploy"}
-: ${TOPOLOGY_FILE="${DEPLOY_DIR}/topology.json.sample"}
-: ${TESTNAME="$(basename ${0})"}
+: "${TEST_DIR="$(realpath "$(dirname "${0}")")"}"
+: "${BASE_DIR="${TEST_DIR}/../.."}"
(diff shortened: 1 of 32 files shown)


obnoxxx:
quite a patchset @jarrpa! gonna try it out

should we hold the block/object pods as different prs until the actual functionality gets merged?

jarrpa:
This depends on the block PR because I didn't feel like holding the block and object PRs further.

obnoxxx:
Not only depends on #285 (deploy gluster-block-provisioner) but also depends on #299 (deploy s3).

obnoxxx:
added do-not-merge. since we need to merge the other PRs first...

obnoxxx:
Or can we rework the patches to push the test-framework enhancements first?

jarrpa:
The main problem was https://github.com/gluster/gluster-kubernetes/pull/302/commits/383b2f85ea3d6fd5a216c5e113d4a07d353fd6b9 from #305, which depends on the rollback feature. I just moved it in here, calling it part of the test suite enhancement (since it kind of is).

obnoxxx:
The thing is, if you would just put these small patches up as separate PRs, I would keep merging them. Also please put into the commit msg *why* you are doing stuff (e.g. why you start ntpd).

And why would that patch rely on the rollback feature?
I think this is wrong. It does not rely on it. But if you move it before the intro of rollback, you need to remove the hunk in rollback.sh and instead adapt the intro of rollback to include restart of ntpd ....
That's the usual business if you move patches past one another ;-)

jarrpa:
Ugh, fine ;P

jarrpa:
...actually, no. The vagrant changes for ntp only came up after I found an issue by running the enhanced test suite. There's no reason to move the ntp patch before the introduction of rollback, as there is no need for it then. I'm amending the commit message but leaving it where it is.

jarrpa:
@obnoxxx Thus pushed.

tests/simple/common/shell_tests.sh
```diff
@@ -0,0 +1,13 @@
...
+	bash -n "${1}"
+}
+
+test_shellcheck() {
+	if ! which shellcheck ; then
+		echo "ShellCheck not found: skipping..."
+		return 0
+	fi
+
+	shellcheck -x -s bash -e SC2181,SC2029,SC1091,SC1090 "${1}"
```
obnoxxx: This sneaks in additional exceptions, which should get into their own conscious commit, imho.

tests/complex/run.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-TEST_DIR="$(realpath $(dirname $0))"
+# shellcheck disable=SC2034
 TESTNAME=""
+TEST_DIR="$(realpath "$(dirname "${0}")")"
```
obnoxxx: Ugh, I did not even know that this works. Do the brackets take precedence over the quotes?

tests/complex/lib.sh
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash
 
 # honor variables set from the caller:
-: ${TEST_DIR:="$(realpath $(dirname $0))"}
-: ${BASE_DIR:="${TEST_DIR}/../.."}
-: ${VAGRANT_DIR:="${BASE_DIR}/vagrant"}
-: ${DEPLOY_DIR:="${BASE_DIR}/deploy"}
-: ${TOPOLOGY_FILE:="${DEPLOY_DIR}/topology.json.sample"}
-: ${TESTNAME:="$(basename $0)"}
+: "${TEST_DIR="$(realpath "$(dirname "${0}")")"}"
```
obnoxxx: this seems to be a shellcheck fix

tests/complex/lib.sh
```diff
@@ -1,12 +1,13 @@
...
+: "${TEST_LOG="${TEST_DIR}/gk-tests.log"}"
```
obnoxxx: this introduces a new variable

tests/complex/lib.sh
```diff
@@ -1,12 +1,13 @@
...
+: "${BASE_DIR="${TEST_DIR}/../.."}"
```
obnoxxx: shellcheck fix?

tests/complex/lib.sh
```diff
@@ -1,12 +1,13 @@
...
+: "${VAGRANT_DIR="${BASE_DIR}/vagrant"}"
```
obnoxxx: quoting/shellcheck?

obnoxxx:
This patchset seems to consist of two patchsets actually:
1. The patchset that factors out the shell syntax checks and runs them on all shell scripts (and fixes the scripts).
2. The patchset with extends the complex tests.

Could you please separate the shell-check patches into a separate PR? moving that would also make the patchsets smaller and more obvious and remove the current interweavement of the two sets of patches (e.g. see the comments to the "Honor inherited variables if set" patch.)

tests/complex/lib.sh
```diff
@@ -84,10 +84,14 @@ copy_deploy() {
 }
 
 pull_docker_image() {
-	local image=$1
-	cd ${VAGRANT_DIR}
-	for NODE in node0 node1 node2 ; do
-		ssh -F ${SSH_CONFIG} ${NODE} "sudo docker pull ${image}"
+	cd "${VAGRANT_DIR}" || exit 1
+
+	local image=${1}
```
obnoxxx: quotes?

tests/complex/lib.sh
```diff
@@ -84,10 +84,14 @@ copy_deploy() {
...
+	local vstatus
+
+	vstatus=$(vagrant status | grep "node" | awk '{print $1}')
```
obnoxxx: quotes?

tests/simple/common/shell_tests.sh
jarrpa: Yes, brackets denote subshells and bash processes all contents within as raw bash (so you don't have to e.g. escape quotes) which also means it has to be valid/shellcheck-correct syntax on its own.

tests/complex/lib.sh
jarrpa: Oops, missed that one.

tests/complex/lib.sh
jarrpa: Not needed.

jarrpa:
I refuse to split this into two PRs. It would be more trouble than its worth to go back and remove the shellcheck-related additions to fixes introduced prior to expanding shellcheck's coverage. I don't want to, for example, have two patches for the same line of this sort:

```
-: ${TESTNAME:="$(basename $0)"}
+: ${TESTNAME="$(basename $0)"}

...

-: ${TESTNAME="$(basename $0)"}
+: ${TESTNAME="$(basename "${0}")"}
```

Doing so would be ridiculous, so I just included any shellcheck-related changes into the actual fix.

Regarding your comments on https://github.com/gluster/gluster-kubernetes/pull/302/commits/f5536e5d7c2c136fa28c8f55b0e387e378787cc7 : I didn't just add quotes I removed the `:` from the bash variable substitution. The `:` made it so that the lib variables did not honor previous values of variables if they were empty strings.

obnoxxx:
@jarrpa I don't think it would be ridiculous, but exactly the right thing to do. Just so much less confusing.
Also, it blends nicely with my request to split this in two patchsets: one with the shell check patches, and then one which work on the complex tests.

I am very adamant about my point that if possible a patch should not do more than one thing. If that means a line should be touched by two or more patches in a row: fine! :-)

jarrpa:
@obnoxxx And I am very adamant in NOT subjecting developers to extreme levels of busywork for the sake of pendanticness. I am updating the commit messages to include notes of where I updated for proper syntax as well. You can split the patches and submit your own PRs if you like, I am not splitting up the work to that degree.

obnoxxx:
@jarrpa it's just the result of me trying to understand and properly review the patches with the due diligence.
The problem is putting too many not directly related patches into one PR, which leads to unclean patches.
I am interested in good resulting patches. And patches that mix unrelated things are not good imho, even if the overall resulting code is good. If you submit a large patchset, one quite legitimate ask is to group the patches together logically. And that means in this case to dissect some changes out of one patch and into another one. You may call it pedanticness. I call it due diligence to come up with good patches that people can make sense of in hindsight. If you're proposing a big patchset, I think you should be willing to spend some amount of time on patch hygiene,  in massaging the patches even if the resulting code remains unchanged. Compared to the actual writing of the code, this should be just a tiny fraction of busywork.

jarrpa:
@obnoxxx I did already spend quite some time on patch hygiene, in some cases to the test of my patience. I believe that this current state plus a few commit msg comments will suffice for anyone to understand what happened on an individual patch basis.

jarrpa:
@obnoxxx In the interest of getting things done, I relented on your bikeshedding. The quotation fixes have been moved out of the fix for honoring variable values.

obnoxxx:
Thanks a lot @jarrpa! This ordering and patch layout looks very clean now. Also note that I was not bikeshedding but requesting what I thought was needed to complete the work on this great patchset!

I only need to do a final review on the "overhaul test framework" patch, before I'm ready to merge.

tests/complex/run.sh
```diff
@@ -5,12 +5,10 @@ TESTNAME=""
...
 source "${TEST_DIR}/lib.sh"
 
-${TEST_DIR}/test-setup.sh || fail
+run -e "${TEST_DIR}/test-setup.sh"
 
-${TEST_DIR}/test-gk-deploy.sh || fail
+run -e "${TEST_DIR}/test-gk-deploy.sh"
 
-${TEST_DIR}/test-dynamic-provisioning.sh || fail
+run "${TEST_DIR}/test-dynamic-provisioning.sh"
```
obnoxxx: why no `-e` here?

tests/complex/test-gk-deploy.sh
```diff
@@ -4,10 +4,9 @@ TEST_DIR="$(realpath $(dirname $0))"
...
 source "${TEST_DIR}/lib.sh"
 
-ssh_config || fail "ERROR to creating ssh-config"
+copy_deploy
 
-copy_deploy || fail "ERROR copying deployment files"
+run -r master -e "${TEST_DIR}/test-inside-gk-deploy.sh" "Test basic deployment"
 
-run_on_node "${TEST_DIR}/test-inside-gk-deploy.sh" master || fail
+run -r master "${TEST_DIR}/test-inside-gk-deploy.sh" "Test basic deployment idempotence"
```
obnoxxx: why no `-e` here

tests/complex/test-dynamic-provisioning.sh
```diff
@@ -5,11 +5,6 @@ DOCKER_IMAGE="gcr.io/google_containers/nginx-slim:0.8"
...
 
-ssh_config || fail "ERROR creating ssh config"
+pull_docker_image "${DOCKER_IMAGE}"
 
-pull_docker_image "${DOCKER_IMAGE}" || fail "ERROR pulling nginx docker images"
-
-run_on_node "${TEST_DIR}/test-inside-dynamic-provisioning.sh" master || fail
-
-pass
 run -r master "${TEST_DIR}/test-inside-dynamic-provisioning.sh" "Test dynamic provisioning"
```
obnoxxx: this line must have sneaked into an earlier commit. `run` did not exist before...

obnoxxx: It sneaked in in commit "tests/complex: Remove single use of LIB_DIR in tests"

obnoxxx:
Ok, I see a little bit of why with the current run() we don't run the helper functions: It assumes that the command is a script / file, currently functions won't work. Of course functions would only work locally, too.

obnoxxx:
Still it might be worthwile to extend run to be able to locally cope with functiouns.
(I might do it in a subsequent patch.)
Main motivation: I'd like to keep the simple functions like `create_vagrant` etc clean of error treatment. This is why I had it on the callers explicitly before, and "run" would be a nice, centralized way to do it consistently.

jarrpa:
- On it.
- The main purpose of `-e` is to prevent the test suite from continuing if a test failure would mean all subsequent tests would fail. While it is the case that dynamic provisioning test failing would mean other tests are going to fail, it's not a guarantee. On the other hand, if gk-deploy fails then we don't have an environment to test further in. :)
- I'd rather not. The error treatment is in the functions because it is basically always going to be the same. It's needless work to require manual error checking, and similarly needless to wrap them in run() and have to supply an error message every time. (Also, run() uses a subshell and we want these commands run in the same shell)

obnoxxx:
ok. fair point.

obnoxxx:
So I'm only waiting for the update that removes the early use of "run"  from the patch that predates its introduction and then I'll be good to go

jarrpa:
Update pushed.

obnoxxx:
LGTM. Thanks for your patience, @jarrpa, that's quite stunning work.

Issue #298, opened 2017-08-15
 Couldn't find an alternative telinit implementation to spawn.


inspired-geek:
Hello! I'm having issue with freshly installed k8s cluster

```
# kubectl --version
Kubernetes v1.7.3
```
 on 4 servers with latest CentOS

```
# cat /etc/redhat-release
CentOS Linux release 7.3.1611 (Core)
```

with docker CE 17.05-0

```
# docker -v
Docker version 17.05.0-ce, build 89658be
```

After pull repository I've set up topology.json and run `.\dk-deploy -g` in `deploy` folder.

```
# ./gk-deploy -g
Welcome to the deployment tool for GlusterFS on Kubernetes and OpenShift.

Before getting started, this script has some requirements of the execution
environment and of the container platform that you should verify.

The client machine that will run this script must have:
 * Administrative access to an existing Kubernetes or OpenShift cluster
 * Access to a python interpreter 'python'
 * Access to the heketi client 'heketi-cli'

Each of the nodes that will host GlusterFS must also have appropriate firewall
rules for the required GlusterFS ports:
 * 2222  - sshd (if running GlusterFS in a pod)
 * 24007 - GlusterFS Daemon
 * 24008 - GlusterFS Management
 * 49152 to 49251 - Each brick for every volume on the host requires its own
   port. For every new brick, one new port will be used starting at 49152. We
   recommend a default range of 49152-49251 on each host, though you can adjust
   this to fit your needs.

For systems with SELinux, the following settings need to be considered:
 * virt_sandbox_use_fusefs should be enabled on each node to allow writing to
   remote GlusterFS volumes

In addition, for an OpenShift deployment you must:
 * Have 'cluster_admin' role on the administrative account doing the deployment
 * Add the 'default' and 'router' Service Accounts to the 'privileged' SCC
 * Have a router deployed that is configured to allow apps to access services
   running in the cluster

Do you wish to proceed with deployment?

[Y]es, [N]o? [Default: Y]:
Using Kubernetes CLI.
NAME      STATUS    AGE
default   Active    15h
Using namespace "default".
Checking that heketi pod is not running ... OK
serviceaccount "heketi-service-account" created
clusterrolebinding "heketi-sa-view" created
node "kub66" labeled
node "kub67" labeled
node "kub68" labeled
node "kub69" labeled
daemonset "glusterfs" created
Waiting for GlusterFS pods to start ...
```
and stuck with it

In k8s dashboard I can see deployed daemon set
[...]

PR #304, opened 2017-08-19
support docker 17.06 and kubernetes 1.7.2

hpgood:
fix bug of:https://github.com/gluster/gluster-kubernetes/issues/298

### DIFF
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -23,6 +23,12 @@ spec:
       - image: gluster/gluster-centos:latest
         imagePullPolicy: IfNotPresent
         name: glusterfs
+        env:
+        - name: SYSTEMD_IGNORE_CHROOT
+          value: "1"
+        command:
+        - /usr/lib/systemd/systemd
+        - --system
         volumeMounts:
         - name: glusterfs-heketi
           mountPath: "/var/lib/heketi"
@@ -52,22 +58,16 @@ spec:
         readinessProbe:
           timeoutSeconds: 3
           initialDelaySeconds: 40
-          exec:
-            command:
-            - "/bin/bash"
-            - "-c"
-            - systemctl status glusterd.service
+          tcpSocket:
+            port: 24007
           periodSeconds: 25
           successThreshold: 1
           failureThreshold: 15
         livenessProbe:
           timeoutSeconds: 3
           initialDelaySeconds: 40
-          exec:
-            command:
-            - "/bin/bash"
-            - "-c"
-            - systemctl status glusterd.service
+          tcpSocket:
+            port: 24007
           periodSeconds: 25
           successThreshold: 1
           failureThreshold: 15


deploy/kube-templates/glusterfs-daemonset.yaml
```diff
@@ -23,6 +23,12 @@ spec:
       - image: gluster/gluster-centos:latest
         imagePullPolicy: IfNotPresent
         name: glusterfs
+        env:
+        - name: SYSTEMD_IGNORE_CHROOT
+          value: "1"
+        command:
+        - /usr/lib/systemd/systemd
+        - --system
```
humblec: This is NOT recommended from `systemd` in production mode.  so, please dont merge.

obnoxxx: @humblec where is this insight from? At least in the regular manpages and docs I found, there is no such statement. Not saying you are wrong, but I'm curious to see references.

humblec: @obnoxxx here it is :)
```
--snip-- from man page

       --system, --user
           For --system, tell systemd to run a system instance, even if the process ID is not 1, i.e. systemd is not run as init process.  --user does the opposite, running a user
           instance even if the process ID is 1. Normally, it should not be necessary to pass these options, as systemd automatically detects the mode it is started in. These options
           are hence of little use except for debugging. Note that it is not supported booting and maintaining a full system with systemd running in --system mode, but PID not 1. In
           practice, passing --system explicitly is only useful in conjunction with --test.

--/snip--
```

obnoxxx: Hmm, thanks, I must have read over this. ;-)

It also does not seem to make sense to me (at least it's not justified in the manpage), and I think it must probably be revisited with containers not giving out PID1 necessarily...

jarrpa: No I didn't. That doesn't say it's a bad idea, just that it isn't supported. Different things. :)

hpgood:
I think we donnot real need full systemd ,but a part fun of systemd which we need.docker also is not a full copy of host ,but it can support the min env for app to run.and docker and google_container/pause do lots of job which systemd has to do.so I think systemd should support optional args to make min self(just like systemd --system),donot do the jobs which docker or google/pause had done.so --system arg tag not only just for test,should change to work fine more and more with docker and google/pause.now is container times.

jarrpa:
@hpgood That's all well and good, but we still need testing, verification, and preferably insight. :) The "right" thing to do would be to figure out how split the GlusterFS container into several containers together in a pod, but that effort will take significantly longer.

hpgood:
@jarrpa At the first time,i was try to splite the GlusterFS contanier into two container together in a pod ,but have some error(maybe need spent long time to resolve),the glusterd proccess cannot running.so I was try to using `systemd --system` ,and it run success.

deploy/kube-templates/glusterfs-daemonset.yaml
poettering: so, we actually officially supported running systemd as non-PID1 once upon a time, but removed that pretty early on due to bitrot and not having a clear usecase. Supporting that isn't particularly hard to do, running systemd as PID 2 should mostly work (except for C-A-D handling/SIGPWR and things like that) in theory. However, the code is sprinkled with "getpid() == 1" checks to distuingish the modes of operation. If there's a clear usecase, a clear user and somebody stepping up to do the work we could restore this in systemd and officially support it. If so, please file an issue/PR on the system github project. But other than that: it's unlikely to work properly if you run systemd as PID 2 but insist on system mode right now.

Lennart

obnoxxx: @poettering Thanks for clarification, Lennart. This makes a lot of sense. We'll discuss and possibly come up with a PR/issue/further discussion. 
- Michael

jarrpa: @poettering Ah, good to know. Thanks! :)

PR #305, opened 2017-08-19
Some minor fixes

jarrpa:
A few minor fixes found in testing #302.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/305)


obnoxxx:
@jarrpa, this PR does not look like "Some minor fixes": It contains new features like deploying the s3 and gluster-block-provisioiner containers...

obnoxxx:
OK, it is said to depend on #302, which itself depends on #285 and #299.
Hence the do-not-merge...

Can we re-order the patches so that we can merge these first? (as well as the actual re-working of the test framework of #302)

jarrpa:
Updated to include the patches from the gluster-s3 branch that this was dependent on.

deploy/gk-deploy
```diff
@@ -256,6 +256,9 @@ check() {
   if [[ ${cond} =~ $number_re ]]; then
     wait_limit=${cond}
   fi
+  if [[ "${resource}" == *s ]]; then
```
obnoxxx: This is a little weird at first sight.
* Firstly, does it mean you change how the second argument is used, depending on whether the first argument ends with the letter 's'? This sounds a little fragile, or implicit.
* And at the very least it should be explained in the commit message, and maybe a function header comment.
* Should `*s` be quoted? Like `"*s"`?

jarrpa: It is weird, I agree, but I couldn't think of a better way to implement it without making the argument parsing (of a helper function!) even more convoluted than it already is. And I just realized I didn't document this function in an earlier patch like I meant to, so I'll do that here.

Also, for some reason globs in bash tests don't get quoted. I don't entirely understand why, but I think it breaks if you quote them.

obnoxxx: Mainly concerned that this different semantics depending whether the first parameter ends with an 's' or not seems fragile and surprising to me, from the perspective of a user of this function.

jarrpa: Again, I agree. My previous comment still stands.

obnoxxx:
This breaks test.
One problem we see which was covered by the tests before is that we see `sed: can't read heketi.json.template: No such file or directory` because gk-deploy is called from a different directory.. i have a fix for that which i'm going to propose separately.
But the real problem is that the check namespace goes into a loop (which takes quite some time) and then eventually gk-deploy reports

> Using namespace "invalid".

Instead of the expected

> Namespace 'invalid' not found.

==> this needs to be fixed

jarrpa:
Fixed the faulty detection of invalid namespace. It takes a bit because I set the timeout to 10 seconds to give Kube some time to finish coming up. We could consider putting in some extra logic in check() to verify that Kubernetes is up first? But I'd ask for that as an RFE.

obnoxxx:
@jarrpa could you consider #319 and possibly merge and rebase on top of it? This test success in travis at least is a false positive. Running test locally now to verify...

jarrpa:
Updated to include a fix for a bug we found in `tests/simple/gk-deploy/stubs`. Also added a few more minor fixes, like not leaving behind `heketi.json`s. :)

jarrpa:
Updated to include a bug fix in the gk-deploy simple test. Also, I opted to go with your recent suggestion of normalizing paths with `cd+dirname+pwd`, as I found a couple error cases in running the complex tests.

obnoxxx:
Thanks @jarrpa, this looks great now!
Running a final local complex test run before merging...

obnoxxx:
works. we really need to run the complex tests in some CI :-)

PR #310, opened 2017-08-21
tests:complex: use vagrant caching.

obnoxxx:

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/310)

### DIFF
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -37,7 +37,7 @@ fail() {
 
 create_vagrant() {
 	cd ${VAGRANT_DIR}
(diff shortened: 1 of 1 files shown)


jarrpa:
I think this should be its own test and not run by default. We can't guarantee that the home directory of someone manually running this will have enough space to host the yum cache (or that the user wants that clutter in their home directory).

obnoxxx:
@jarrpa that's a fair point. I can still export VAGRANT_CACHE=1 before running the tests if I wish to use.

PR #319, opened 2017-08-22
Fix selftests in CI by removing realpath

obnoxxx:
Realpath is not available in travis CI.
The bad thing is that these errors got unnoticed.
* So this removes the uses of realpath.
* And it adds a test to make sure that realpath is not used...
* Finally, (this could be put into a separate PR), this fixes the simple testsuite to do proper error propagation. (Previously, errors could go unnoticed even if the test itself detected it.)

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/319)

### DIFF
--- a/tests/complex/lib.sh
+++ b/tests/complex/lib.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # honor variables set from the caller:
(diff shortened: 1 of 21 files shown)


obnoxxx:
Finally, all ok in travis, no false passes. :-)

@jarrpa ptal

PR #329, opened 2017-08-30
Don't try to forward terminal to `kubectl exec` calls

YorikSar:
If run not from terminal, kubectl shows this error:

  Unable to use a TTY - input is not a terminal or the right kind of file

It makes the script think that topology loading failed and exit with
error:

  Error loading the cluster topology.

That said we don't need to forward terminal here at all - we call CLI
that doesn't interact with terminal.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/329)


jarrpa:
Man, I am still so new to Kube. :)

LGTM!

PR #330, opened 2017-08-30
Don't add stderr to stdout when parsing `kubectl get pod` output

YorikSar:
Due to issue in kubectl ( kubernetes/kubernetes#48924 ) it outputs some
warnings to stderr and they are considered to be output lines and it
breaks the script in 2 places.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/330)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -298,7 +298,7 @@ check() {
       break
     fi
     sleep 2
-    res=$(${CLI} get "${resource}" --no-headers --show-all "${select}" 2> /dev/null)
+    res=$(${CLI} get "${resource}" --no-headers --show-all "${select}" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


jarrpa:
The second change for `heketi_pod` looks like it makes sense. However, for the first one I'd as that you redirect stderr output to `/dev/null` instead. We don't care about any error messages, we'll just wait for the timeout if we don't reach a satisfactory condition.

YorikSar:
@jarrpa  I've added that redirection.

jarrpa:
A curiosity: Is the space between the angle bracket and the destination valid? Even so, for stylistic consistency please remove it. Yes, I'm picky. ;)

YorikSar:
@jarrpa Yes, it's valid. Will remove...

jarrpa:
Thanks! LGTM

PR #335, opened 2017-08-31
Added new env. variable GB_GLFS_LRU_COUNT for indicating maximum number

SaravanaStorageNetwork:
Added new env. variable GB_GLFS_LRU_COUNT for indicating maximum number
of block hosting volumes.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/335)

### DIFF
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -23,6 +23,9 @@ spec:
       - image: gluster/gluster-centos:latest
         imagePullPolicy: IfNotPresent
         name: glusterfs
+        env:
+        - name: GB_GLFS_LRU_COUNT
+          value: 15
         resources:
           requests:
             memory: 100Mi
--- a/deploy/ocp-templates/glusterfs-template.yaml
+++ b/deploy/ocp-templates/glusterfs-template.yaml
@@ -36,6 +36,9 @@ objects:
         - image: gluster/gluster-centos:latest
           imagePullPolicy: IfNotPresent
           name: glusterfs
+          env:
+          - name: GB_GLFS_LRU_COUNT
+            value: "${GB_GLFS_LRU_COUNT}"
           resources:
             requests:
               memory: 100Mi
@@ -128,3 +131,8 @@ parameters:
   displayName: Daemonset Storage Node Label
   description: This is the value that will be used for the storagenode label in the daemonset node selector.
   value: glusterfs
+- name: GB_GLFS_LRU_COUNT
+  displayName: Maximum number of block hosting volumes
+  description: This value is to set maximum number of block hosting volumes.
+  value: "15"
+  required: true


obnoxxx:
Hi @SaravanaStorageNetwork , thanks for the patch!
I am not quite sure what this is really supposed to do.
First of all, this variable name seems to be used in gluster-block in the init scripts / systemd unit files.
And best I can tell, it has nothing to do with the number of block hosting volumes.

Please clarify!

SaravanaStorageNetwork:
@obnoxxx  This specific variable is passed as an argument to gluster-blockd.

The current value is 5 which enables to create 5+1 block hosting volumes ( Refer https://bugzilla.redhat.com/show_bug.cgi?id=1478046).

This value is small and we wish this value to be configurable. 
As this value is passed as an argument in the template (hence to the container), even if the pod is restarted the value will persist.

obnoxxx:
@SaravanaStorageNetwork ok, so at least the commit message is confusing and misleading:
This no control for the number of block-hosting volumes that can be created, but it is a control for **something** obscure (to me) that (according to a certain BZ which does not explain but mere states it) lets us create $VALUE + 1 block-hosting volumes **safely**...

obnoxxx:
Ok, @SaravanaStorageNetwork, this is not your fault. The Bug description is also bad and does not really explain, and likewise the seeting name in gluster-blockd... ;-) This is a completely counter-intuitive hack, imho. Why doesn't someone fix the memleaks in glfs_fini? :-)

pkalever:
@SaravanaStorageNetwork 
So the value of GB_GLFS_LRU_COUNT set here gain priority over the Env variable set in the systemd unit ?

#cat /usr/lib/systemd/system/gluster-blockd.service
[...]
Environment="GB_GLFS_LRU_COUNT=5"
[...]
ExecStart=/usr/local/sbin/gluster-blockd --glfs-lru-count $GB_GLFS_LRU_COUNT --log-level $GB_LOG_LEVEL $GB_EXTRA_ARGS
[...]

Have we tested this ?

humblec:
@pkalever, its not about priority, rather its about what this variable supposed to or the intention?

pkalever:
@humblec what is the point of setting this value here, if the one which is in systemd unit overrides it ?

So I just wanted to check who gains here ?

humblec:
@pkalever I am pointing @obnoxxx's question.

obnoxxx:
@humblec thanks for pointing out the direction of my question. I would like this whole thing clarified. and ideally cleaned.

But I do actually have the same question as @pkalever as well: Is this effective? And this is the question about precedence.

SaravanaStorageNetwork:
@pkalever @obnoxxx  This is verified. 
A script (during docker container startup) is updating this parameter present in systemd script.

PS: You cannot access docker/openshift env. variables from inside systemd.

SaravanaStorageNetwork:
@pkalever  We should preferably add all env. variables in env. file (which is all commented now) - I will raise a issue(https://github.com/gluster/gluster-block/issues/36)  gluster-block - please check this later.

jarrpa:
@SaravanaStorageNetwork If this goes through, you also need a patch for the kube-templates. :)

SaravanaStorageNetwork:
> If this goes through, you also need a patch for the kube-templates. :)

@jarrpa  Sure - will take care of this once this is merged 👍

jarrpa:
@SaravanaStorageNetwork Please just add it to this one, since it's the same change and functionality.

deploy/kube-templates/glusterfs-daemonset.yaml
```diff
@@ -104,3 +107,8 @@ spec:
       - name: glusterfs-ssl
         hostPath:
           path: "/etc/ssl"
+parameters:
+- name: GB_GLFS_LRU_COUNT
+  displayName: Maximum number of block hosting volumes
+  description: This value is to set maximum number of block hosting volumes.
+  value: 15
```
jarrpa: No `parameters` for DaemonSets.

jarrpa:
`kube-templates/glusterfs-daemonset.yaml` should still have the change, it's just that there's no such thing as template parameters to configure, we do that with `sed` lines in gk-deploy. For now, though, we can simplify this by just setting the env var to 15 and we can make it configurable from gk-deploy later.

SaravanaStorageNetwork:
done

jarrpa:
You misunderstood my earlier comment, I only wanted the `kube-templates` file changed. Please ensure the following:

* `deploy/kube-templates/glusterfs-daemonset.yaml` should list the variable and a hard-coded value of `15`.
* `deploy/ocp-templates/glusterfs-daemonset.yaml` should list the variable with a `"${GB_GLFS_LRU_COUNT}"` (note the quotes) parameter, and there should also be a parameter entry at the bottom with the default value of `"15"` (note the quotes)

Also, please squash all your changes into one commit then do a `git push -f` to overwrite the branch. We don't need to see everything fixed for the patch, just the final patch. :)

jarrpa:
@SaravanaStorageNetwork You missed the quotes around the `15` at the bottom of the template. :(

jarrpa:
Ah! I didn't get a notification for the push, it seems.

LGTM!

PR #336, opened 2017-09-01
vagrant: make it a little more easy to use the created local registry

obnoxxx:
Have the commented out value match the ip address that we use
for the vagrant boxes, so that in case we use docker-registry-run.sh
to create the registry, we just need to comment out the line
to make use of it.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/336)


jarrpa:
Does this work by default? The localhost gets assign `.1` on the subnet of the private net?

obnoxxx:
yes. on my systems (libvirt) always

jarrpa:
Cool, LGTM.

PR #339, opened 2017-09-02
add --overwrite for nodes

pronix:
require if label already exists

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/339)


jarrpa:
Oh! Good catch. :)

LGTM

PR #351, opened 2017-09-11
when using ssh executor provide /etc/fstab as ENV

raghavendra-talur:
If ssh executor is used, it also generally means that Gluster is not
containerized. In such instances, we don't have the startup script in
Gluster container which would look at /var/lib/heketi/fstab for bricks.
Hence, this patch changes the HEKETI_FSTAB ENV to /etc/fstab when
sshexec is chosen.

For now, this is sufficient. When and if we want to support ssh access
to Gluster containers, we will have to provide a flag --fstab to let
users decide where the file should reside. Not now.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/351)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -34,6 +34,7 @@ EXISTS_DEPLOY_HEKETI=0
 EXISTS_HEKETI=0
 EXISTS_OBJECT=0
 EXECUTOR="kubernetes"
+FSTAB="/var/lib/heketi/fstab"
 SSH_KEYFILE="/dev/null"
 SSH_USER="root"
 SSH_SUDO="false"
@@ -393,6 +394,7 @@ while [[ $# -ge 1 ]]; do
         ;;
         s*|-ssh-keyfile)
         EXECUTOR="ssh"
+        FSTAB="/etc/fstab"
         SSH_KEYFILE=$(assign "${key:${keypos}}" "${2}")
         if [[ $? -eq 2 ]]; then shift; fi
         keypos=$keylen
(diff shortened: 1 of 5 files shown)


humblec:
LGTM

raghavendra-talur:
`>You also need to edit the kube and ocp template files to take the variables. Currently this does nothing to change the value used when the resources are created.
`
Thanks! sent in too much of a hurry.

`>Also, how does this actually manifest in the container? Wouldn't the /etc/fstab file in the container be destroyed when the container is?
`
In container , we change the fstab path to /var/lib/heketi/fstab which is bind mounted because the whole of /var/lib/heketi is.

@jarrpa I don't need to change the heketi.json.template , do I?

jarrpa:
>> Also, how does this actually manifest in the container? Wouldn't the /etc/fstab file in the container be destroyed when the container is?
> In container , we change the fstab path to /var/lib/heketi/fstab which is bind mounted because the whole of /var/lib/heketi is.

Right. So in this case, we won't be bind mounting anything into the container there, right?

raghavendra-talur:
> Also, how does this actually manifest in the container? Wouldn't the /etc/fstab file in the container be destroyed when the container is?
> In container , we change the fstab path to /var/lib/heketi/fstab which is bind mounted because the whole of /var/lib/heketi is.
> Right. So in this case, we won't be bind mounting anything into the container there, right?

I don't understand the question here.

In a setup where gluster is not containerized but heketi is, heketi populates the /etc/fstab file in gluster node thereby ensuring that node reboots mount the bricks.

raghavendra-talur:
@jarrpa Also, let me know if you want the patches to be squashed.

deploy/gk-deploy
```diff
@@ -751,14 +753,14 @@ if [[ ${GLUSTER} -eq 1 ]] && [[ ${EXISTS_GLUSTERFS} -eq 0 ]] && [[ ${EXISTS_HEKE
 fi
 
 if [[ ${EXISTS_DEPLOY_HEKETI} -eq 0 ]] && [[ ${EXISTS_HEKETI} -eq 0 ]]; then
-  sed -e "s/\${HEKETI_EXECUTOR}/${EXECUTOR}/" -e "s/\${SSH_PORT}/${SSH_PORT}/" -e "s/\${SSH_USER}/${SSH_USER}/" -e "s/\${SSH_SUDO}/${SSH_SUDO}/" "${SCRIPT_DIR}/heketi.json.template" > heketi.json
+  sed -e "s/\${HEKETI_EXECUTOR}/${EXECUTOR}/" -e "s/\${HEKETI_FSTAB}/${FSTAB}/" -e "s/\${SSH_PORT}/${SSH_PORT}/" -e "s/\${SSH_USER}/${SSH_USER}/" -e "s/\${SSH_SUDO}/${SSH_SUDO}/" "${SCRIPT_DIR}/heketi.json.template" > heketi.json
```
humblec: I think above path resolution is wrong ie HEKETI_FSTAB/FSTAB

humblec: ignore above, I see its conf k:v pair not the path. Thanks.

raghavendra-talur:
I don't get the spellcheck error on my machine that is failing the build here. However, I have sent a patch to make changes as suggested by spellcheck in #352

PR #352, opened 2017-09-11
fix SC2199 in tests/simple/gk-deploy/stubs/cli.sh

raghavendra-talur:
As per information provided in
https://github.com/koalaman/shellcheck/wiki/SC2199, we need to use $*
instead of ${@} in cli.sh

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/352)


raghavendra-talur:
@jarrpa done.

Issue #198, opened 2017-03-13
Verify topology json format


SaravanaStorageNetwork:
Verify topology json file which is provided as input for gk-deploy and error out if any errors in json format.

PR #364, opened 2017-09-29
Check json format of topology.json

SaravanaStorageNetwork:
Check json format of topology.json

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/364)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -569,13 +569,21 @@ import json
 import argparse
 
 file = open('${TOPOLOGY}', 'r')
-topo = json.load(file)
+try:
+  topo = json.load(file)
+except ValueError as e:
+  print("Invalid json format in : %s Error: %s" % (file, e))
+  sys.exit(1)
 
 for cluster in topo['clusters']:
   for node in cluster['nodes']:
     print(str(node['node']['hostnames']['manage'][0]))
 END
 )
+  if [[ $NODES == "Invalid json format"* ]]; then
+    echo -e "$NODES \nFix json format and rerun"
+    exit 1
+  fi
 fi
 
 if [[ "x${CLI}" == "x" ]]; then


SaravanaStorageNetwork:
sample run with invalid topology file

```
$ ./deploy/gk-deploy -g /usr/share/heketi/topology-sample.json 
Welcome to the deployment tool for GlusterFS on Kubernetes and OpenShift.

Before getting started, this script has some requirements of the execution
environment and of the container platform that you should verify.

The client machine that will run this script must have:
 * Administrative access to an existing Kubernetes or OpenShift cluster
 * Access to a python interpreter 'python'

Each of the nodes that will host GlusterFS must also have appropriate firewall
rules for the required GlusterFS ports:
 * 2222  - sshd (if running GlusterFS in a pod)
 * 24007 - GlusterFS Management
 * 24008 - GlusterFS RDMA
 * 49152 to 49251 - Each brick for every volume on the host requires its own
   port. For every new brick, one new port will be used starting at 49152. We
   recommend a default range of 49152-49251 on each host, though you can adjust
   this to fit your needs.

The following kernel modules must be loaded:
 * dm_snapshot
 * dm_mirror
 * dm_thin_pool

For systems with SELinux, the following settings need to be considered:
 * virt_sandbox_use_fusefs should be enabled on each node to allow writing to
   remote GlusterFS volumes

In addition, for an OpenShift deployment you must:
 * Have 'cluster_admin' role on the administrative account doing the deployment
 * Add the 'default' and 'router' Service Accounts to the 'privileged' SCC
 * Have a router deployed that is configured to allow apps to access services
   running in the cluster

Do you wish to proceed with deployment?

[Y]es, [N]o? [Default: Y]: Y
Invalid json format in : <open file '/usr/share/heketi/topology-sample.json', mode 'r' at 0x7f5cfa88f780> Error: Expecting , delimiter: line 4 column 20 (char 48) 
Fix json format and rerun

```

SaravanaStorageNetwork:
Fixes #198 
As [heketi](https://github.com/heketi/heketi) can process topology file separately, will add JSON format check there as well.

SaravanaStorageNetwork:
@jarrpa done

PR #373, opened 2017-10-23
vagrant: use kubernetes version 1.7.8

phlogistonjohn:
Without specifying the Kubernetes version the playbooks fail due to
behavior changes in kubeadm and friends.
Explicitly ask for 1.7.8 as that is the newest version of 1.7 in the
google yum repo.

Signed-off-by: John Mulligan <jmulligan@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/373)

### DIFF
--- a/vagrant/global_vars.yml
+++ b/vagrant/global_vars.yml
@@ -1,5 +1,5 @@
 ---
-kubernetes_version: 1.6.1
+kubernetes_version: 1.7.8
 kubernetes_token: abcdef.1234567890abcdef
 install_pkgs:
 - wget
@@ -12,9 +12,9 @@ install_pkgs:
 - iptables-utils
 - iptables-services
 - docker
-- kubeadm
-- kubelet
-- kubectl
+- kubeadm-1.7.8
+- kubelet-1.7.8
+- kubectl-1.7.8
 - ntp
 # The following variables control the use and configuration of a custom docker
 # registry:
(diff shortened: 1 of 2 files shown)


vagrant/global_vars.yml
```diff
@@ -12,9 +12,9 @@ install_pkgs:
 - iptables-utils
 - iptables-services
 - docker
-- kubeadm
-- kubelet
-- kubectl
+- kubeadm-1.7.8
+- kubelet-1.7.8
+- kubectl-1.7.8
```
jarrpa: What happens if you don't specify the version explicitly here? Would the latest RPMs get pulled? If so, is that bad?

I guess my main thought is that the "latest" RPMs will always be there, whereas I don't know what the retention policy of older RPMs is. Not sure if breaking because of a version change (like we just did) or breaking because the RPMs are gone are really any different, though. :)

phlogistonjohn: Yes, you end up with the latest RPMs - which are as I write this 1.8.1. This version changes some behaviors and command output in a couple of places and this breaks the current playbooks.
As of this moment the quickstart guide can't run to completion because of that.

I'd like to follow up on this change later with some that make it usable with1.8.1 and possibly use the version variable in global_vars.yml to determine what version of the packages are used and conditionalize behaviors.
However, in the short term I hoped to keep the scope small.

jarrpa: I say we can let the scope creep up just a little bit: replace the other instances of 1.7.8 with the version variable. :)

jarrpa:
LGTM!

PR #389, opened 2017-11-11
fix pods ready condition

caogj:
in my cluster settings, where `istio-initializer` https://istio.io/` is enabled, which will auto inject `istio-proxy` container into any pod in kubernetes. So the right condition  should be '2/2'.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/389)

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -294,7 +294,7 @@ assign() {
 #   The following resource types have unique conditional states:
 #
 #   * Pods
-#     - default: state reports '1/1'
+#     - default: state reports the desired number of containers are ready, eg. '1/1'.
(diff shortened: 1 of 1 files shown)


deploy/gk-deploy
```diff
@@ -294,7 +294,7 @@ assign() {
 #   The following resource types have unique conditional states:
 #
 #   * Pods
-#     - default: state reports '1/1'
+#     - default: state reports 'ready/desired' where `ready == desired`, eg. '1/1'.
```
jarrpa: Change this to: `the desired number of containers are ready.`

caogj: done! thanks.

jarrpa:
LGTM!

PR #397, opened 2017-11-23
vagrant: ./up.sh should use --no-parallel to prevent download issues

nixpanic:
When I run ./up.sh in a clean environment, at least one of the VMs does
not come up. The following error is then displayed:

    An error occurred while executing multiple actions in parallel.
    Any errors that occurred are shown below.

    An error occurred while executing the action on the 'node0'
    machine. Please handle this error then try again:

    The box 'centos/7' could not be found or
    could not be accessed in the remote catalog. If this is a private
    box on HashiCorp's Atlas, please verify you're logged in via
    `vagrant login`. Also, please double-check the name. The expanded
    URL and error message are shown below:

    URL: https://atlas.hashicorp.com/centos/7
    Error: Could not resolve host: atlas.hashicorp.com

By using `vagrant up --no-parallel ...` the box gets downloaded for the
first VM and the others are able to use it as well.

Suggested-by: Raghavendra Talur <rtalur@redhat.com>
Signed-off-by: Niels de Vos <ndevos@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/397)

### DIFF
--- a/vagrant/up.sh
+++ b/vagrant/up.sh
@@ -1,5 +1,13 @@
 #!/bin/sh
 
+# download the box(es) to prevent issues with parallel "vagrant up"
+BOXES=$(awk -F '"' '/config.vm.box/ {print $2}' Vagrantfile | sort -u)
+for BOX in ${BOXES}; do
+  if ! (vagrant boxes list | grep -q -e "^${BOX}[[:space:]]"); then
+    vagrant box add "${BOX}"
+  fi
+done
+
 export ANSIBLE_TIMEOUT=60
 vagrant up --no-provision "${@}" \
     && vagrant provision


raghavendra-talur:
LGTM

jarrpa:
I just tried to test this and did not run into the error described. What version of vagrant are you running?

nixpanic:
I am using Vagrant that comes with Fedora 27: vagrant-1.9.1-3.fc27.noarch

Running `vagrant box add ...` or similar can be done. For me the problem
is *very* reproducible, I only have to delete the boxes and libvirt image.

jarrpa:
Hmm... I can't reproduce it even with removing the libvirt image. Regardless, adding something like this at the top of the `up.sh` script seems to be working for me:
```
BOXES=$(cat Vagrantfile | grep config.vm.box | cut -d "\"" -f 2)

for box in $BOXES; do
  if [[ "x$(vagrant box list | grep $box)" == "x" ]]; then
    vagrant box add "$box"
  fi
done
```
This will prompt the user for the provider they want to use, but should only happen once. The `BOXES` variable is in case we want to specify different boxes for the VMs for cross-platform testing. Possibly a bit of over-engineering, so take that into consideration. :)

obnoxxx:
Yeah, strange - I never saw that either...

Just tried on a freshly upgraded system. removed boxes and all libvirt uploaded copies.
ran `./up.sh`. All other boxes wait until the first has downloaded the box, and afterwards go parallel...

obnoxxx:
In order to thest the new @centos-ci integration:

ok to test

obnoxxx:
retest this please

obnoxxx:
@centos-ci retest this please

nixpanic:
@centos-ci retest this please

nixpanic:
Oh, nice, it seems that a manually updated Jenkins job shows the progress - https://i.imgur.com/czdUvbo.png

Obviously the 'changes requested' will still need to be addressed, the retest was for the Jenkins job.

nixpanic:
@purpleidea I'm trying to automatically download the `confg.vm.box`, but Vagrant wants to know what provider (libvirt in my case) to use and shows me a selection on the console. Latest commit 98b28b35de0c924db9fa5f140b4916bbb39ea5f3 contains a description of the approach. The prompt for a provider will prevent any CI to work without extra hacks. Any idea?

purpleidea:
@nixpanic You need to set the provider, see here: https://ttboj.wordpress.com/2013/12/09/vagrant-on-fedora-with-libvirt/

purpleidea:
@nixpanic Just saw your commit saying provider options aren't working. Did you set the box options?
eg: https://github.com/purpleidea/oh-my-vagrant/blob/master/vagrant/Vagrantfile#L1036
and
https://github.com/purpleidea/oh-my-vagrant/blob/master/vagrant/Vagrantfile#L1090

box and box url ?

nixpanic:
Also, it seems that the `--provider=libvirt` works if it is passed as last argument on the commandline. Maybe a combination of the following can be made to work:

```
vagrant status master | awk ... # fetch the provider
vagrant box add centos/7 --provider=${PROVIDER}
```

nixpanic:
With a little help from @purpleidea (thanks!), I have found the cause for the actual problem that prevented the VMs from starting. The network configuration on my laptop re-initializes when new networks are connected (to work with multiple VPNs). Because Vagrant starts new (libvirt) networks, and network devices for each VM, systemd detects a 'too many restarts' problem and disables the dnsmasq service. No DNS means no name resolution, and this seems to cause failures while downloading the box (strangely all other connections stay alive).

Instead of fixing the problem in gluster-kubernetes, I've changed the script that configures DNS to skip libvirt managed networks/devices: nixpanic/nm-separate-dns@f2e9621112130a41a3eda86fda9cbd2ea998c381

PR #401, opened 2017-11-24
a few improvements on complex test runner scripts

obnoxxx:
I have some more in the making, but sending out these first, since they are ready and self-contained.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/401)


jarrpa:
Tests still ran well, LGTM!

Issue #392, opened 2017-11-13
Error waiting for job 'heketi-storage-copy-job' to complete.


olegsmetanin:
I have deployed fresh Kubernetes cluster (1 master + 2 nodes) without GlusterFS. On each node I run
```
wipefs -a /dev/vdb
modprobe dm_thin_pool
apt-get install -y glusterfs-client
```
create topology.json with Private IP
```
cat <<EOF > topology.json
{
  "clusters": [
    {
      "nodes": [
        {
          "node": {
            "hostnames": {
              "manage": [
                "master01"
              ],
              "storage": [
                "10.3.35.73"
              ]
            },
            "zone": 1
          },
          "devices": [
            "/dev/vdb"
          ]
        },
        {
          "node": {
            "hostnames": {
              "manage": [
                "node01"
              ],
              "storage": [
                "10.4.103.199"
              ]
            },
            "zone": 1
          },
          "devices": [
            "/dev/vdb"
          ]
        },
        {
          "node": {
            "hostnames": {
              "manage": [
                "node02"
              ],
              "storage": [
                "10.2.150.9"
              ]
            },
            "zone": 1
          },
          "devices": [
            "/dev/vdb"
          ]
        }
      ]
    }
  ]
}
EOF
```

After that I run ./gk-deploy -gvy
```
./gk-deploy -gvy
Using Kubernetes CLI.

Checking status of namespace matching 'default':
default   Active    20m
Using namespace "default".
Checking for pre-existing resources...
  GlusterFS pods ...
Checking status of pods matching '--selector=glusterfs=pod':

Timed out waiting for pods matching '--selector=glusterfs=pod'.
not found.
  deploy-heketi pod ...
Checking status of pods matching '--selector=deploy-heketi=pod':

Timed out waiting for pods matching '--selector=deploy-heketi=pod'.
not found.
  heketi pod ...
Checking status of pods matching '--selector=heketi=pod':

Timed out waiting for pods matching '--selector=heketi=pod'.
not found.
  gluster-s3 pod ...
Checking status of pods matching '--selector=glusterfs=s3-pod':

Timed out waiting for pods matching '--selector=glusterfs=s3-pod'.
not found.
Creating initial resources ... /usr/local/bin/kubectl -n default create -f /root/gluster-kubernetes/deploy/kube-templates/heketi-service-account.yaml 2>&1
serviceaccount "heketi-service-account" created
[...]

PR #409, opened 2017-12-22
Added version matching requirement to docs

AnatolyRugalev:
See issue #392.

By default on Ubuntu 16.04 version of glusterfs-client do not match version of server which causes client crash on mount attempt.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/409)

### DIFF
--- a/docs/setup-guide.md
+++ b/docs/setup-guide.md
@@ -32,6 +32,10 @@ requirements:
  * Each node requires that the `mount.glusterfs` command is available. Under
   all Red Hat-based OSes this command is provided by the `glusterfs-fuse`
   package.
+  
+  * GlusterFS client version installed on nodes should be as close as possible
+   to the version of the server. To get installed versions run
+   `glusterfs --version` or `kubectl exec <pod> -- glusterfs --version`.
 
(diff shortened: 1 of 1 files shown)


docs/setup-guide.md
```diff
@@ -32,6 +32,10 @@ requirements:
  * Each node requires that the `mount.glusterfs` command is available. Under
   all Red Hat-based OSes this command is provided by the `glusterfs-fuse`
   package.
+  
+  * GlusterFS client version installed on nodes should match version of the
+   server which is running inside container. To get installed versions run
```
jarrpa: Very close. A more accurate statement would be `should be as close as possible to the version of the server`. If the client is more advanced than the server, for instance, things should still work.

docs/setup-guide.md
```diff
@@ -32,6 +32,10 @@ requirements:
...
+   `glusterfs --version` and `kubectl exec <pod> -- glusterfs --version`.
```
jarrpa: s/and/or/

jarrpa:
@AnatolyRugalev Sorry for the long silence, I don't get notifications for pushes to the PR. But, LGTM!

PR #415, opened 2018-01-10
vagrant: Update kubeadm join command

jarrpa:
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/415)

### DIFF
--- a/vagrant/roles/nodes/tasks/main.yml
+++ b/vagrant/roles/nodes/tasks/main.yml
@@ -45,7 +45,7 @@
 
 - block:
   - name: kubeadm join with master
-    command: kubeadm join --skip-preflight-checks --token={{ kubernetes_token }} {{ hostvars['master'].ansible_eth1.ipv4.address }}:6443
+    command: kubeadm join --discovery-token-unsafe-skip-ca-verification --ignore-preflight-errors=all --token={{ kubernetes_token }} {{ hostvars['master'].ansible_eth1.ipv4.address }}:6443
 
   - name: create root kube dir on node
     file:


phlogistonjohn:
I think this adding this option will prevent running with older versions of k8s. Perhaps it can be conditional on the version of kubernetes? Or if we don't want to lock ourselves to version numbers maybe "scrape" the output of the help to see if our version supports that switch?

jarrpa:
@phlogistonjohn Ah, fair point. I'll do the conditional version check.

Man, we should really move to a proper cycle of releases... ;)

jarrpa:
@phlogistonjohn I put in a conditional check and verified that it works for latest. However, when I tried to set `kubernetes_version: 1.8.0` in `global_vars.yml` I couldn't even get to the point of testing the conditional because it failed to find any kube RPMs ending in `-1.8.0`. This leads me to ask: does setting `kubernetes_version` to anything other than latest actually work? If not, I would want this PR to merge and we can figure out a proper, broader solution later.

phlogistonjohn:
I recently (this week) ran it with 1.8.6 when I was testing something else out. However, I did run into an additional issue using that version. Yum couldn't figure out the correct version of kubernetes-cni to use so I had to manually add kubernetes-cni-0.5.1-1.x86_64 to the install_pkgs list.

I'll leave it to you to decide if that extra hurdle means taking the PR now or deferring it.

jarrpa:
Hmm... yeah, I'll opt to take the PR now. Without this any setups that try to use `latest` kube will not succeed.

jarrpa:
retest this please

jarrpa:
retest this please

jarrpa:
Looks like there's a libvirt error in the CI. Ignoring it for now.

PR #416, opened 2018-01-10
wait for kubernetes before trying to deploy weave

phlogistonjohn:
This fixes an issue I was hitting with mutltiple runs. Usually the
first run of the playbooks would work fine but if a different issue
was hit and I neded to re-run them I kept hitting errors on
the weave step related to connection failures. This change adds a
wait for the port to be available before running that troublesome
step.

Signed-off-by: John Mulligan <jmulligan@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/416)

### DIFF
--- a/vagrant/roles/master/tasks/main.yml
+++ b/vagrant/roles/master/tasks/main.yml
@@ -64,6 +64,13 @@
       group: vagrant
   when: not kubelet.stat.exists
 
+- name: wait for k8s to be ready
+  wait_for:
+    host: localhost
+    port: 6443
+    state: started
+    timeout: 60
+
 - name: create weave network
   command: kubectl apply -f https://git.io/weave-kube-1.6
 


jarrpa:
I don't understand... why would you need to re-run them if they ran successfully?

phlogistonjohn:
The fist time the weave steps works but something else fails, say for example an unrelated command on one of the nodes. Then on subsequent runs the command to set up weave fails. I see that my commit message isn't very clear. Do you think I should try and reword it?

jarrpa:
Oh, I see. I've run into that as well, but also in cases when Kube hasn't even had a chance to be set up yet, thus there would be no point in waiting on the port. The right thing to do would be to check if weave is already deployed and skip if so... but I haven't had/taken the time to figure out how to do that yet. At present I just destroy the whole thing and restart.

phlogistonjohn:
Well, the kubectl command is smart enough not to do any extra work if weave is deployed, but only if it can connect to the server. :-)
I do occasionally throw the whole thing away when things aren't working but even with caching it's not super-fast so I try to avoid doing it if I can.

jarrpa:
In that case my question becomes: if we got far enough to deploy weave the first time that means kubelet is up and running, so why on the second pass do we need to wait for it to come up if it should be already running? We don't even re-run kubeadm if the config exists, so what's causing kubectl to fail?

phlogistonjohn:
AFAICT kubelet gets restarted during the plays. This is from 'systemctl status kubelet' added right before the probe I added for this PR:

```
● kubelet.service - kubelet: The Kubernetes Node Agent
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/kubelet.service.d
           └─10-kubeadm.conf
   Active: active (running) since Wed 2018-01-10 18:13:58 UTC; 3s ago
     Docs: http://kubernetes.io/docs/
 Main PID: 14133 (kubelet)
   CGroup: /system.slice/kubelet.service
           └─14133 /usr/bin/kubelet --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true --network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin --cluster-dns=10.96.0.10 --cluster-domain=cluster.local --authorization-mode=Webhook --client-ca-file=/etc/kubernetes/pki/ca.crt --cgroup-driver=systemd
```

I think that the initial play does enough "extra stuff" that it delays the weave setup until after kubelet is fully initialized. In these subsequent runs that has already been done and the kubectl apply command is run much sooner after kubelet is restarted. 

If your reaction is that we shouldn't restart kubelet, I'm cool with that but I haven't yet figured out when/why that happens.

phlogistonjohn:
I don't know why I overlooked this earlier:
```
- name: enable kube services
  service: name={{ item }} state=restarted enabled=yes daemon_reload=yes
  with_items:
    - docker
    - kubelet
    - ntpd
```

I checked the ansible docs to verify my belief and state=restarted does indeed always restart the services.

jarrpa:
The only scenario I can think of where it makes sense to leave that as `restarted` is that it then allows us to modify configuration that requires a restart and then automate the restart. That might be valuable enough on its own.

All right, I'll take it! LGTM.

PR #423, opened 2018-01-21
vagrant: configure VMs with 2BG RAM

nixpanic:
I used the VMs for building custom @heketi containers, maybe that also caused some additional memory problems. Except for messages in `dmesg`, none of the logs suggested OOM troubles, it was annoying to find the cause.

1GB seems a little on the low end, but 2GB might be a little much in case others have less powerful systems to develop on.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/423)

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -25,7 +25,7 @@ Vagrant.configure("2") do |config|
             vb.cpus = 2
         end
         master.vm.provider :libvirt do |lv|
-            lv.memory = 1024
+            lv.memory = 2048
             lv.cpus = 2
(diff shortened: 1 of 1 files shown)


jarrpa:
Hmm... yeah, that's hard. I think I'm leaning towards leaving it at 1GB by default? If only because the main use cases for this environment are: 1.) running our test suite and 2.) enabling users to try out the project. In both cases I would want as small a memory footprint as possible.

nixpanic:
1GB RAM for the testsuite is enough. If people like me experiment with heavier usage, they should be prepared to debug problems and increase the RAM for their usecase.

PR #431, opened 2018-01-29
increase failure threshold for glusterfs pods

raghavendra-talur:
It is possible that lvm takes more time to bring up the brick devices
if there are many number of bricks. Increase the threshold to account
for the same.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/431)


jarrpa:
ok to test

jarrpa:
Right, still that libvirt error... LGTM!

PR #441, opened 2018-02-16
vagrant: use 'host-passthrough' for CPU in the VMs

nixpanic:
On occasion the vagrant tests fail in the CentOS CI with the following
error:

  Call to virDomainCreateWithFlags failed: the CPU is incompatible with
  host CPU: Host CPU does not provide required features: svm

See-also: https://bugzilla.redhat.com/show_bug.cgi?id=1467599
See-also: https://bugzilla.redhat.com/show_bug.cgi?id=1386223#c10
See-also: vagrant-libvirt/vagrant-libvirt#667
Signed-off-by: Niels de Vos <ndevos@redhat.com>


obnoxxx:
ok to test

PR #444, opened 2018-02-21
Rename ocp-templates as openshift-templates

SaravanaStorageNetwork:


jarrpa:
I intentionally avoided doing this because I wanted to keep the directory names short. Given that we are going to be getting rid of these in the near future, I don't want to introduce needless changes.

Issue #75, opened 2016-12-14
gk-deploy script should fail to execute if the topology.json file not found


MohamedAshiqrh:
Currently gk-deploy is started even if the topology file is not specified. It should verify and fail if not found.

Issue #440, opened 2018-02-16
Gluster pods should be able to load certain lvm/device-mapper modules


nixpanic:
There is no guarantee that the `dm_thin_pool`, `dm_mirror` and `dm_snapshot` kernel modules are loaded. Creating volumes with Heketi enablee thin-provisioning for the bricks. This will fail if the device-mapper/lvm modules are not available.

Two solutions are available (that I can think of):

1. add a script like `/etc/sysconfig/modules/*.modules` on the host to load the modules on bootup
2. enable the Gluster pods to load the module if they are missing

Approach 2 looks the most reasonable, and the automatic loading of modules is already standard practice for non-containerized environments. In order to make 2 happen in the most cases, it will be needed to bind-mount `/lib/modules` from the host into the pod. Without this bind-mount, the pod will not be able to run `modprobe` with the right `.ko` files.

See-also: #434 `/etc/sysconfig/modules/*.modules` approach

jarrpa:
I'm also fond of approach 2. This makes it so that the kernel modules aren't loaded at boot if GlusterFS is removed.

PR #457, opened 2018-04-04
Let the glusterfs pod load kernel modules

nixpanic:
In certain cases the required kernel modules for creating thin-provisioned LVM Logical Volumes are not loaded. The lvm tools will cause automatic loading of the modules, but this will fail if the  `.ko` files are not available in the pod where the lvm command (likely `lvcreate`) is executed.

An example of these errors that is happening in the CentOS CI:
```
07:21:38 /usr/bin/kubectl -n default exec -i deploy-heketi-6c7d48f8b-jnx68 -- heketi-cli -s http://localhost:8080 --user admin --secret '' setup-openshift-heketi-storage --listfile=/tmp/heketi-storage.json  2>&1
07:21:53 Error: Unable to execute command on glusterfs-gg9h7:   /usr/sbin/modprobe failed: 1
07:21:53 thin: Required device-mapper target(s) not detected in your kernel.
07:21:53 Run `lvcreate --help' for more information.
07:21:53 command terminated with exit code 255
07:21:53 Failed on setup openshift heketi storage
07:21:53 This may indicate that the storage must be wiped and the GlusterFS nodes must be reset.
```

By bind-mounting `/usr/lib/modules` into the pod, the loading of the kernel modules is now possible.

### DIFF
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -55,6 +55,9 @@ spec:
         - name: glusterfs-ssl
           mountPath: "/etc/ssl"
           readOnly: true
+        - name: kernel-modules
+          mountPath: "/usr/lib/modules"
+          readOnly: true
         securityContext:
           capabilities: {}
           privileged: true
@@ -109,3 +112,6 @@ spec:
       - name: glusterfs-ssl
         hostPath:
           path: "/etc/ssl"
+      - name: kernel-modules
+        hostPath:
+          path: "/usr/lib/modules"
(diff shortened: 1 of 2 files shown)


jarrpa:
I'd like this better if you also incorporated the module loading into the test suite so we can verify it does what it needs to do. Not that the CI is working right now, but I'd at least run it locally.

obnoxxx:
ok to test

nixpanic:
The commit message of the patch mentions `Fixes: #75`. This is incorrect and should have been #440.

The gluster/gluster-containers#75 issue is pretty much the same, but for the plain `Dockerfile` when deploying manually without Kubernetes or OpenShift.

PR #476, opened 2018-05-13
Gluster Operator Design

jarrpa:
A design doc to hopefully get some traction on the development of a Gluster operator. From the introduction:

```
This document aims to settle on a concrete set of functionality that a Gluster
operator should have. It is outside the scope of this document to specify any
implementation details of this functionality. The desired functionality should
not be strictly limited by the extents of current-day technologies but should
not dismiss the costs of waiting for and executing forward development in those
technologies.

In addition, this document only specifies the desired functionality for
declaring a 1.0 release. Further enhancements and feature roadmaps will be
captured elsewhere.
```

### DIFF
--- a/docs/design/operator.md
+++ b/docs/design/operator.md
@@ -1,0 +1,200 @@
+# Gluster Operator 1.0 Design
+
+## Introduction
+
+This document specifies the overall design for a Gluster operator. An operator
+is a Kubernetes [controller](https://github.com/kubernetes/sample-controller)
+that is focused on managing a specific application with the goal of automating
+common tasks that an administrator would typically perform.
+
+This document is a formalization of many hours of discussion by several people.
+Credit and thanks to Humble Devassy Chirammal (@humblec),Steve Watt
+(@wattsteve), and John Strunk (@JohnStrunk) for their early and insightful
+contributions to the discussions.
+
+This document aims to settle on a concrete set of functionality that a Gluster
+operator should have. It is outside the scope of this document to specify any
+implementation details of this functionality. The desired functionality should
+not be strictly limited by the extents of current-day technologies but should
+not dismiss the costs of waiting for and executing forward development in those
+technologies.
+
+In addition, this document only specifies the desired functionality for
+declaring a 1.0 release. Further enhancements and feature roadmaps will be
+captured elsewhere.
+
+### Motivation
+
+The Gluster operator aims to further streamline the experience of deploying and
+maintaining Gluster clusters for use in a Kubernetes environment. It will
+achieve this through the use of common Kubernetes components and paradigms,
+helping further integrate persistent storage into the Kubernetes experience.
+
+## Configurations
+
+The Gluster operator will be capable of managing a number of configurations of
+Gluster clusters. Gluster's flexibility allows it to run in many typs of
+environments and the operator will aim to preserve that flexibility as much as
+possible.
+
+### External Deployments
+
+External deployments are those in which the Gluster servers are not running as
+containers inside Kubernetes. They are running on external, potentially
+dedicated, storage nodes.
+
+### Hosted Deployments
+
+Hosted deployments are those in which the Gluster servers are running as
+containers but their storage devices are not managed by the operator. In these
+deployments, an administrator is required to maintain any underlying storage
+devices and inform the operator as to any changes in their configuration.
+
+### Managed Deployments
+
+Managed deployments are those in which the Gluster servers are running as
+containers and their storage devices are managed by the operator. These
+deployments are designed for cloud environments where storage devices can be
+provisioned programmatically.
+
+## Features
+
+Since the Gluster operator will support a variety of deployment configurations,
+it will not be feasible to provide identical feature sets across all
+configurations. The following sections note which features will be available to
+which configuration.
+
+All operational features will be designed to be as idempotent as possible. It is
+acknowledged that this may not be immediately possible due to restrictions in
+the underlying technology but an effort will still be made.
+
+### Deployment
+
+The Gluster operator will be able to deploy a new Gluster cluster using a set of
+configuration parameters. This configuration will allow for extensive
+customization if desired but will assume sane defaults for as many parameters as
+possible to allow for as minimal a configuration as possible.
+
+**Target Configurations:** Hosted, Managed
+
+### Import
+
+The Gluster operator will be able to import an existing Gluster cluster at any
+stage of initialization. The operator's desired configuration for the cluster
+will need to be compatible with the state of the cluster at the time of import.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Cluster Upgrade
+
+The Gluster operator will be able to execute a rolling upgrade of any hosted or
+managed Gluster cluster. It will trigger an upgrade one node at a time, waiting
+for the node to come up and for the cluster to report no pending heals to any
+files before triggering a subsequent node.
+
+**Target Configurations:** Hosted, Managed
+
+### Monitoring
+
+To provide event-driven features, the Gluster operator will monitor various
+aspects of the clusters it will automate.
+
+#### Storage Capacity
+
+The Gluster operator will monitor the storage capacity of the overall cluster.
+Storage capacity is defined here as the amount of space that can be committed to
+new Gluster volumes and not how much free space is available. It will trigger
+events when storage capacity reaches "Warning" levels and "Critical" levels.
+These levels will be configurable, and should be set to a value that will allow
+for corrective action to take place before the storage reaches maximum capacity.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Volume Density
+
+The Gluster operator will monitor the total volume density of any cluster it is
+automating. It will trigger events when the number of volumes in the cluster
+reaches a pre-defined level that should prompt corrective action. This value
+will be configurable, and should be set to a value that will allow for
+corrective action to take place before cluster nodes lag or crash.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Device Health
+
+The Gluster operator will monitor the health of all devices in an automated
+Gluster cluster. It will report a "Warning" when a device reports failing health
+and will report a "Critical" problem when a device becomes inaccessible. Support
+for this feature may vary form one device type to another.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Volume Health
+
+The Gluster operator will monitor the health of all volumes in an automated
+Gluster cluster. It will report a "Warning" when a volume reports failing bricks
+and will report a "Critical" problem when a volume becomes inaccessible.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Node Health
+
+The Gluster operator will monitor the health of all nodes in an automated
+Gluster cluster. It will report a "Critical" problem when a node becomes
+inaccessible.
+
+**Target Configurations:** External, Hosted, Managed
+
+### Cluster Management
+
+In response to the monitoring events above, the Gluster operator will take any
+number of corrective management actions.
+
+#### Storage Scaling
+
+In the event that the overall cluster reports a critical capacity level, the
+operator will deploy a set number of additional storage nodes with new storage
+devices attached. The number of additional nodes to be deployed will be
+configurable.
+
+**Target Configurations:** Managed
+
+#### Cluster Scaling
+
+In the event that the overall cluster reports a critical volume density level,
+the operator will deploy a new cluster of Gluster nodes to receive volume
+creation requests. The number of additional nodes to be deployed will be
+configurable.
+
+**Target Configurations:** Managed
+
+### Disaster Recovery
+
+In response to the health monitoring events above, the Gluster operator will
+take any number of corrective actions to recover from a situation that could
+result in potential data loss.
+
+#### Brick Replacement
+
+If a volume reports a failing brick, the Gluster operator will try to create a
+new brick and use it as a replacement for the failing one.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Device Replacement
+
+If a node reports a failing device, the Gluster operator will try to provision a
+new device and attach it to the node to replace the failing device. It would
+then recreate any bricks on the failing device and trigger a brick replacement
+for any affected volumes.
+
+**Target Configurations:** Managed
+
+#### Node Replacement
+
+If a node becomes inaccessible, the Gluster operator will try to deploy a new
+Gluster node with new storage devices attached. It would then recreate any
+bricks from the missing node and trigger a brick replacement for any affected
+volumes.
+
+**Target Configurations:** Managed


docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+The Gluster operator aims to further streamline the experience of deploying and
+maintaining Gluster clusters for use in a Kubernetes environment. It will
+achieve this through the use of common Kubernetes components and paradigms,
+helping further integrate persistent storage into the Kubernetes experience.
+
+## Configurations
+
+The Gluster operator will be capable of managing a number of configurations of
+Gluster clusters. Gluster's flexibility allows it to run in many typs of
```
phlogistonjohn: Minor typo

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+and will report a "Critical" problem when a device becomes inaccessible. Support
+for this feature may vary form one device type to another.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Volume Health
+
+The Gluster operator will monitor the health of all volumes in an automated
+Gluster cluster. It will report a "Warning" when a volume reports failing bricks
+and will report a "Critical" problem when a volume becomes inaccessible.
```
vbellur: It would be nice to monitor essential (glusterd, self-heal-daemon etc.) & optional services (quotad, bitrot-daemon) in addition to bricks as part of this activity.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+#### Storage Scaling
+
+In the event that the overall cluster reports a critical capacity level, the
+operator will deploy a set number of additional storage nodes with new storage
+devices attached. The number of additional nodes to be deployed will be
+configurable.
+
+**Target Configurations:** Managed
+
+#### Cluster Scaling
```
JohnStrunk: I think this should be out of scope at this time. GD2 should fix the need for multiple clusters, meaning we'd only need this if it falls short... a good reason to push it out.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+**Target Configurations:** Managed
+
+### Disaster Recovery
+
+In response to the health monitoring events above, the Gluster operator will
+take any number of corrective actions to recover from a situation that could
+result in potential data loss.
+
+#### Brick Replacement
```
JohnStrunk: Per the meeting last month, this will be a Gluster-internal operation, assuming sufficient free capacity is available. The operator is only concerned with adding and removing capacity & nodes from the cluster.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+If a volume reports a failing brick, the Gluster operator will try to create a
+new brick and use it as a replacement for the failing one.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Device Replacement
```
JohnStrunk: As a first step, this can be greatly simplified by having 1 device per gluster node, making a failed device the same as a failed node.
That 1 device could be an arbitrarily sized RAID array if an admin chooses.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+If a node reports a failing device, the Gluster operator will try to provision a
+new device and attach it to the node to replace the failing device. It would
+then recreate any bricks on the failing device and trigger a brick replacement
+for any affected volumes.
+
+**Target Configurations:** Managed
+
+#### Node Replacement
```
JohnStrunk: Here again, there is a split between the operator's tasks and Gluster's internal responsibilities. The operator would create a new node, and once data has been rebuilt, destroy the old node. Gluster would internally handle the brick reconstruction.

docs/design/operator.md
jarrpa: Oh, right! Fair enough. Tying this back to the monitoring section, would it make sense to have some sort of "out of space" event that the operator could react to in addition to the two watermarks?

travisn:
You might consider adding the CRD spec to the design doc. This will help define the desired state the admin can specify when he creates the gluster cluster. It will also help differentiate between what is desired state managed by the operator vs configuration tasks that the operator likely wouldn't manage. 

For the configuration tasks, consider that gluster should have a "toolbox" pod that allows the admin to drop in and perform operational tasks manually. The toolbox allows running tasks that the operator might not be expected to run, either because they are one-time admin tasks or because the admin just needs to override the operator.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Import
+
+The Gluster operator will be able to import an existing Gluster cluster at any
+stage of initialization. The operator's desired configuration for the cluster
+will need to be compatible with the state of the cluster at the time of import.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Cluster Upgrade
```
mrobson: Can the cluster upgrade component be made dynamic enough to also be useful in other situations which require a rolling restart of the Gluster nodes? Effectively, it would handle bringing down Gluster cleanly, waiting for X and ensuring the health before continuing. Beyond the Gluster upgrade, the 2 primary examples of X being, upgrading the underlying nodes and restarting the VMs or Servers for OS Patching.

jarrpa:
> You might consider adding the CRD spec to the design doc. This will help define the desired state the admin can specify when he creates the gluster cluster. It will also help differentiate between what is desired state managed by the operator vs configuration tasks that the operator likely wouldn't manage.

I would think the CRD spec would be too close to the implementation of the operator to merit including in this document. As it stands, the CRD in my prototype has evolved over the course of the implementation as I think things through and try things out.

> For the configuration tasks, consider that gluster should have a "toolbox" pod that allows the admin to drop in and perform operational tasks manually. The toolbox allows running tasks that the operator might not be expected to run, either because they are one-time admin tasks or because the admin just needs to override the operator.

Wouldn't that just be the same as exec'ing into one of the Gluster nodes?

docs/design/operator.md
jarrpa: The operator would not be managing the upgrade of any underlying infrastructure. In the case where a given node identity can not move between hosts, it would be up to the administrator to ensure they upgrade such hosts one at a time and wait for the Gluster cluster's health to restore. The only involvement of the operator in that case would be, at best, to report the state of health.

travisn:
@jarrpa great, if the tools are all available in the gluster nodes, no need for a separate toolbox. 

What I'm looking for is a declaration of desired state that can be controlled by the admin. As is, the doc makes it sound like the operator will be able to automatically maintain the health, grow the cluster, and other actions without any input from an admin. Since you're using a CRD you must have desired state that controls this, I just can't see what it is. If the CRD is too detailed, perhaps an abstraction of the CRD?

docs/design/operator.md
mrobson: Agreed, it would not handle those other tasks, but ideally, it would be able to aid them. Instead of everyone else needing to understand how to cleanly terminate gluster (daemon set / pod or otherwise) and check it's health when it comes back, it would be nice for other users and operators to be able to leverage the gluster operator for that functionality. If the tasks are broken out as discretely, that may be possible. Long and short, that kind of reusability is what I am hoping you take into consideration.

jarrpa:
@travisn Ah, okay. The desired state is only changed by the operator in the case of a managed deployment, where both the services and the storage devices are not tied to specific hosts. Otherwise it is up to the administrator to change the CRD object.

I can see the rational for some general state outline, at least. I'll work on that.

docs/design/operator.md
jarrpa: At present, and best I can see, there's nothing more to bringing down a Gluster pod than any other pod. Kubernetes will take care to do as clean a shutdown as it can, to my understanding. As far as watching to health restoration, would simply exposing the state via a condition field in an object's Status be sufficient?

phlogistonjohn:
Perhaps this is too implementation oriented but I think it would be nice if the ability to make changes to the cluster were individually toggle-able. For example you are working with a "managed" cluster but do not want it to automatically perform Cluster scaling but storage scaling might be ok.

IOW, are the target configurations in the doc more like use cases or something that would actually be specified in the system configuration?

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+Gluster cluster. It will report a "Warning" when a volume reports failing bricks
+and will report a "Critical" problem when a volume becomes inaccessible.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Node Health
+
+The Gluster operator will monitor the health of all nodes in an automated
+Gluster cluster. It will report a "Critical" problem when a node becomes
+inaccessible.
```
sbarre-esit: I'll second @vbellur request to monitor additional services. Currently the readiness probe only checks `glusterd`, while `tcmu-runner` `gluster-block-target` and `gluster-blockd` are all critical to a healthy gluster pod. Perhaps that just needs to be pushed into the readiness probe though?

sbarre-esit: Would this also include the other pods like `glusterblock-provisioner` and `heketi` ?

docs/design/operator.md
sbarre-esit: Besides failing bricks, it would be good to also check for unhealed files, or files in split-brain.

docs/design/operator.md
jarrpa: > There is something more in that the volume with bricks hosted by that pod should be clean before shutting down. I'm not sure if sig lifecycle would have any opinions on how that could be ensured.

That's up to Gluster, not the operator. Kube sends a SIGTERM to the main process on pod delete, in our case some sort of init/systemd, which should take care of shutting things down gracefully.

docs/design/operator.md
jarrpa: > Perhaps that just needs to be pushed into the readiness probe though?

Yes. :)

> Would this also include the other pods like `glusterblock-provisioner` and `heketi` ?

That's implementation-specific. Neither of those two are guaranteed to always be around.

jarrpa:
Updated the PR:

* Added a description of state information that the operator will maintain.
* Removed the "cluster scaling" feature.

docs/design/operator.md
```diff
@@ -77,15 +77,7 @@ possible to allow for as minimal a configuration as possible.
 
 **Target Configurations:** Hosted, Managed
 
-### Import
-
-The Gluster operator will be able to import an existing Gluster cluster at any
-stage of initialization. The operator's desired configuration for the cluster
-will need to be compatible with the state of the cluster at the time of import.
-
-**Target Configurations:** External, Hosted, Managed
```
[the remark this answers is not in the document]
jarrpa: Naw, I just won't hold up a 1.0 for it. It'll either be in a 1.5ish release or developed as an external tool to migrate existing clusters to operator configurations.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
+# Gluster Operator 1.0 Design
+
+## Introduction
+
+This document specifies the overall design for a Gluster operator. An operator
+is a Kubernetes [controller](https://github.com/kubernetes/sample-controller)
+that is focused on managing a specific application with the goal of automating
```
humblec: I dont think we have to specify the `sample controller` here, the reason being no bits about implementation is covered here in the PR. We have to discuss further on the implementation side and come up best suited controller model, there are different framework available to implement an operator ( meta controller , operator sdk, ..etc),  we have to discuss further on it and come up with an implementation doc or PR later.

jarrpa: Find me a better link to showcase what a controller is and I'll happily change it. :)

jarrpa: I've asked around about the operator SDK and it very much doesn't even exist yet. While it sounds like it might be useful, I don't feel like waiting around for it.

Reading on metacontroller, it looks like a utility API for writing controllers. It doesn't sound like it has anything particularly novel, and in some aspects looks more constricting than just writing the controller framework by hand. Given that I've already written a controller base, I'm not interested in exploring metacontroller at this time.

JohnStrunk: > I've asked around about the operator SDK and it very much doesn't even exist yet. While it sounds like it might be useful, I don't feel like waiting around for it.

The SDK is here: https://github.com/operator-framework/operator-sdk
While it is still very actively changing, it is certainly worth looking at as a template. It is also probably going to be useful to be compatible with the lifecycle manager: https://github.com/operator-framework/operator-lifecycle-manager

jarrpa: Okay, more accurately: I was told not to use it yet. :) It effectively does not exist for me until it at least hits alpha.

docs/design/operator.md
obnoxxx: > >  Would this also include the other pods like glusterblock-provisioner and heketi ?
> 
> That's implementation-specific. Neither of those two are guaranteed to always be around.

But I think in the current situation, they would certainly belong to the scope of the operator!

obnoxxx: I think this point is valid though. The document reads as if the operator only cares about gluster nodes. But maybe we need to define what are the "nodes in an automated Gluster cluster". Those will be gluster nodes, but possibly other pods in kubernetes. At least currently it will include heketi and (optionally) a gluster-block-provisioner pod. Etc.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
+# Gluster Operator 1.0 Design
```
obnoxxx: Not sure if 1.0 is really appropriate here. It sounds premature and might set false expectations.

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+
+This document aims to settle on a concrete set of functionality that a Gluster
+operator should have. It is outside the scope of this document to specify any
+implementation details of this functionality. The desired functionality should
+not be strictly limited by the extents of current-day technologies but should
+not dismiss the costs of waiting for and executing forward development in those
+technologies.
+
+In addition, this document only specifies the desired functionality for
+declaring a 1.0 release. Further enhancements and feature roadmaps will be
```
obnoxxx: I'd rather say something along the lines of "...specifies a desired minimal set of functionalities for getting started with an operator" ... (again the 1.0 comment).

docs/design/operator.md
```diff
@@ -0,0 +1,200 @@
...
+reaches a pre-defined level that should prompt corrective action. This value
+will be configurable, and should be set to a value that will allow for
+corrective action to take place before cluster nodes lag or crash.
+
+**Target Configurations:** External, Hosted, Managed
+
+#### Device Health
+
+The Gluster operator will monitor the health of all devices in an automated
+Gluster cluster. It will report a "Warning" when a device reports failing health
```
obnoxxx: I think we need to define the term "automated Gluster cluster" in this document. (see below comments)

jarrpa:
Pushed updates:

* Changed desired version from `1.0` to `0.1`.
* Removed the notion of "automated Gluster clusters", I already assume elsewhere that when I say "Gluster cluster" the reader knows I'm talking about those Gluster clusters that are deployed and automated by the operator.

PR #484, opened 2018-06-11
Automatically create a dynamically provisioning StorageClass

patrickmslatteryvt:
Add option to automatically create a dynamically provisioning StorageClass after the GlusterFS daemon-set deployment has completed.

To automatically create a StorageClass named: `glusterfs-storage`

```
./gk-deploy --yes --deploy-gluster --namespace glusterfs --create-dynamic-sc
```

To automatically create a StorageClass named: `glusterfs-ssd-storage`

```
./gk-deploy --yes --deploy-gluster --namespace glusterfs --create-dynamic-sc --dynamic-sc "glusterfs-ssd-storage"
```

I don't have access to an OpenShift cluster to test this against so I didn't include the template for OpenShift.

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -49,6 +49,8 @@ OBJ_USER=""
 OBJ_PASSWORD=""
 OBJ_STORAGE_CLASS="glusterfs-for-s3"
 OBJ_CAPACITY="2Gi"
+CREATE_DYNAMIC_SC=0
+DYNAMIC_STORAGE_CLASS="glusterfs-storage"
 
 usage() {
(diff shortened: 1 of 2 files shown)


patrickmslatteryvt:
OK, sure. I can make those changes.

patrickmslatteryvt:
OK, I messed up on the last one, I get what you mean now. I'll fix it in a day or two, I need to make another test cluster to validate it on first...

Also I was thinking I could add an option to set the StorageClass as default, your thoughts?

jarrpa:
Couldn't hurt to add it. `--storage-class-default` or something like that.

ansiwen:
How about a --storage-class-volumetype option as well? For minikube I would need volumetype: "none", and it would be nice to use this new feature.

PR #486, opened 2018-06-14
Add script to deploy into a single node minikube instance

ansiwen:


ansiwen:
@jarrpa: I'm nor sure it makes sense to make the size of the bricks file configurable. It is a sparse file of 10GB and sits inside an image of an hard-coded size of 13GB, that is created by minikube. I can't think of a scenario where it is useful to make it smaller or bigger. Do you have one?

Creating the ISO was actually the biggest effort for this PR. :sweat_smile: I added some background on it in the commit notes and how to create it by yourself.

jarrpa:
....ah, okay! In that case could you just provide hints on how to create the minikube image as comments in the script? I can almost guarantee that people won't think to look in the commit notes.

ansiwen:
@jarrpa I pushed a new version which only deploys on an existing minikube deployment, since the latest version of minikube includes the heketi/gluster dependencies now in the latest release, after this had been merged: https://github.com/kubernetes/minikube/pull/2925

PR #492, opened 2018-06-26
Update README.md

jayunit100:


README.md
```diff
@@ -78,6 +78,8 @@ $ vagrant ssh master
 [vagrant@master]$ mv topology.json.sample topology.json
 ```
 
+*If not running in the provided vagrant, make sure to edit the topology.json node, zone, device names to match you're kubernetes cluster.*
```
phlogistonjohn: * Please fix the contraction "you're" (you are) to the possessive "your".
* I would also expand "vagrant" to "default Vagrant environment" because its actually fairly easy to tweak the vagrant setup (I do). 
* Kubernetes should be capitalized, I think.
* Your commit message should probably say something like "Update README.md to remind users to match topology.json to vagrant" or something like that

PR #502, opened 2018-07-25
remove show-all as it is depreciated as part of oc get <resources>

SaravanaStorageNetwork:

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -340,7 +340,7 @@ check() {
       break
     fi
     sleep 2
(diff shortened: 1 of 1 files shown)


SaravanaStorageNetwork:
@humblec  PTAL

humblec:
It looks good to me.  @jarrpa @obnoxxx PTAL . This address https://bugzilla.redhat.com/show_bug.cgi?id=1600077

sd2k:
FYI: when using `kubectl` this causes the gk-deploy script to hang at this stage:

```
Checking status of pods matching '--selector=job-name=heketi-storage-copy-job':
```

since the pod completes almost immediately and doesn't appear without `--show-all`.

PR #511, opened 2018-08-24
gk_deploy: add --show-all switch to kubectl call

crunis:
Inside gk_deploy script, in check() function definition, when calling kubectl to check if one of the jobs has finished, the call is missing "--show-all" so it also shows jobs that are in `Completed` status, otherwise if the job has finished it is not listed and the script keeps waiting, eventually timing out.

Sample output:
```
$ kubectl get pods --no-headers -l job-name=heketi-storage-copy-job
No resources found, use --show-all to see completed objects.
$ kubectl get pods --show-all --no-headers -l job-name=heketi-storage-copy-job
heketi-storage-copy-job-8nfz7   0/1       Completed   0         19m
```

### DIFF
--- a/deploy/gk-deploy
+++ b/deploy/gk-deploy
@@ -340,7 +340,7 @@ check() {
       break
     fi
     sleep 2
-    res=$(${CLI} get "${resource}" --no-headers "${select}" 2>/dev/null)
+    res=$(${CLI} get "${resource}" --no-headers --show-all "${select}" 2>/dev/null)
     if [[ ${s} -ne 0 ]] && [[ ${VERBOSE} -eq 1 ]]; then
       reslines=$(echo "$res" | wc -l)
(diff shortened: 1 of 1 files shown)


nixpanic:
ok to test

obnoxxx:
This undoes half of PR #502 ...

@SaravanaStorageNetwork @jarrpa what to do here?

obnoxxx:
Just want to see whether the @centos-ci had not run when triggered by @nixpanic or if it just does not show up in the list of checks:

ok to test

bjwschaap:
Please note that `--show-all` shouldn't be used anymore, since it is deprecated. The default behaviour of `kubectl` in current version is equivalent to the previous show all. See: https://github.com/kubernetes/kubernetes/issues/67895

PR #523, opened 2018-10-02
[RFC] deploy: if available, use new probe script in the glusterfs container image

phlogistonjohn:
If the probe script does not exist the statement falls back to the
previous method of checking only the status of glusterd.

This should only be merged if https://github.com/gluster/gluster-containers/pull/105 is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/523)

### DIFF
--- a/deploy/ocp-templates/glusterfs-template.yaml
+++ b/deploy/ocp-templates/glusterfs-template.yaml
@@ -37,6 +37,8 @@ objects:
           imagePullPolicy: IfNotPresent
           name: glusterfs
           env:
+          - name: GLUSTER_BLOCKD_STATUS_PROBE_ENABLE
+            value: "${GLUSTER_BLOCKD_STATUS_PROBE_ENABLE}"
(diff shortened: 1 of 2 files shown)


deploy/ocp-templates/glusterfs-template.yaml
```diff
@@ -37,6 +37,10 @@ objects:
           imagePullPolicy: IfNotPresent
           name: glusterfs
           env:
+          # set GLUSTER_BLOCKD_STATUS_PROBE_ENABLE to "1" so the
+          # readiness/liveness probe validate gluster-blockd as well
+          - name: GLUSTER_BLOCKD_STATUS_PROBE_ENABLE
+            value: "1"
```
SaravanaStorageNetwork: All parameters are defined below - but missing this variable.

SaravanaStorageNetwork: Also, the default value in the parameters section. (check other variables for reference)

PR #524, opened 2018-10-03
more reliable passing of the complex tests

nixpanic:
New versions of Kubernetes' kubeadm do not have the
--skip-preflight-check option anymore. It has been replaced with
--ignore-preflight-errors. While testing, it seems this flag is not
needed anymore, tests can continue without it.

However, the updated Kubernetes versions seem more strict about the
order of objects in the .yaml files. There are two instances where the
Service refers to a Pod that is defined later on. This seems to prevent
the Service from getting created. Moving the Service definition to the
end, and tests pass again.

See-also: https://github.com/kubernetes/kubernetes/commit/3a0aa06fc9cb78738ba542db6154f71f5d1b92ea

### DIFF
--- a/deploy/kube-templates/deploy-heketi-deployment.yaml
+++ b/deploy/kube-templates/deploy-heketi-deployment.yaml
@@ -1,21 +1,4 @@
 ---
-kind: Service
-apiVersion: v1
-metadata:
-  name: deploy-heketi
-  labels:
-    glusterfs: heketi-service
-    deploy-heketi: service
-  annotations:
-    description: Exposes Heketi Service
-spec:
-  selector:
-    deploy-heketi: pod
-  ports:
-  - name: deploy-heketi
-    port: 8080
-    targetPort: 8080
----
 kind: Deployment
 apiVersion: extensions/v1beta1
 metadata:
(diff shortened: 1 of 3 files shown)


jarrpa:
ok to test
add to whitelist

phlogistonjohn:
@nixpanic I did and yes that was probably the reason for the failure. Since I was under time pressure to test I just used an older version of k8s by editing the global_vars file.
Did you test with any older versions, say one version back from latest?

nixpanic:
On Thu, Oct 04, 2018 at 02:34:50PM +0000, John Mulligan wrote:
> @nixpanic I did and yes that was probably the reason for the failure. Since I was under time pressure to test I just used an older version of k8s by editing the global_vars file.
> Did you test with any older versions, say one version back from latest?

No, I did not try with an older version... If that is a feature we want
to support, it'll get a little more difficult to implement.

jarrpa:
We've generally tried to keep things somewhat backwards-compatible, especially for being able to switch between latest and latest-1. There are some spots where we do version checking that you could use to see if it wouldn't be too bad.

obnoxxx:
So what's the take on this?
Do we want to require backward-compatibility in the templates?

PR #545, opened 2018-12-27
gluster-block needs to dynamically create /dev/uio* devices

nixpanic:
By not bind-mounting /dev in the container anymore, gluster-block fails to create block-volumes. Unfortunately bind-mounting /dev is not possible with CRI-O, so we came up with a different approach to reach the same goal.

The gluster/gluster-container#115 change introduces a feature in the early start script for the glusterfs-server container. It is now possible to `mount --rbind` an alternative /dev from the host onto the /dev path in the container.

It also has been noticed that CRI-O secures `/sys` more than Docker does. tcmu-runner (part of gluster-block) needs write access to `/sys/module` and `/sys/class`.

With these two changes, gluster-block functions again on recent versions of OpenShift, both with Docker and CRI-O runtimes.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/545)

### DIFF
--- a/deploy/kube-templates/glusterfs-daemonset.yaml
+++ b/deploy/kube-templates/glusterfs-daemonset.yaml
@@ -24,6 +24,9 @@ spec:
         imagePullPolicy: IfNotPresent
         name: glusterfs
         env:
+        # alternative for /dev volumeMount to enable access to *all* devices
+        - name: HOST_DEV_DIR
+          value: "/mnt/host-dev"
         # set GLUSTER_BLOCKD_STATUS_PROBE_ENABLE to "1" so the
         # readiness/liveness probe validate gluster-blockd as well
         - name: GLUSTER_BLOCKD_STATUS_PROBE_ENABLE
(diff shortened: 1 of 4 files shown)


nixpanic:
ok to test

nixpanic:
add to whitelist

nixpanic:
In addition to the required changes for gluster-block, two patches have been included so that the CentOS CI works again. @jarrpa @obnoxxx let me know if I should move those out into their own PR.

nixpanic:
On Mon, Jan 07, 2019 at 05:30:38PM +0000, Jose A. Rivera wrote:
> jarrpa requested changes on this pull request.
> 
> The CentOS CI patches should be moved to their own PR, yes. :)

Done. This PR is now rebased on top of
https://github.com/gluster/gluster-kubernetes/pull/551 which needs to be
merged first.

jarrpa:
retest this please

PR #576, opened 2019-04-05
@kubectl@ --show-all option is defaulted to true and deprecated since 1.11 and removed since 1.14

fbarco:
* deploy/gk-deploy : remove @--show-all@ option to @kubectl@ command line

` ~$ kubectl version`
`Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:53:57Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}`
`Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:45:25Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}`

`~$ kubectl get pods --show-all --no-headers -l job-name=heketi-storage-copy-job`
 `Error: unknown flag: --show-all`
....
`Use "kubectl options" for a list of global command-line options (applies to all commands).`

`unknown flag: --show-all`

PR #587, opened 2019-05-02
Removes unknown flag in K8S 1.13

cduverne:
Hello,

This PR updates the gk-deploy script to work with Kubernetes v1.13, where "show-all" is unknown.

Hope this will help.
Cheers


nixpanic:
This looks like a duplicate of #576.

@jarrpa should we still have support for older versions that do support `--show-all` with `kubectl`?

PR #598, opened 2019-06-16
Update gk-deploy

eliezerchavez:
kubectl -n storage get pod --no-headers --show-all --selector="heketi" | awk '{print $1}'                                                                                    
Error: unknown flag: --show-all

This issue doesn't allow the setup completes successfully.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/598)


oruchreis:
I created pods in a custom namespace other than default. So I've changed the line with "--all-namespaces". And I'm getting ip:port from endpoints because show-all is not supported anymore:
```
kubectl get endpoints --no-headers -o wide --all-namespaces | grep 'heketi ' | awk '{print $2}'
```

PR #605, opened 2019-07-18
make heketi admin and user keys mandatory

raghavendra-talur:
Heketi has changed its default mode to authenticated and won't start if
the config file OR env does not have them set. gk-deploy should fail
early if keys are not provided than failing later.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/605)


jarrpa (approved):
LGTM

phlogistonjohn (commented):
looks good to me

Issue #618, opened 2019-10-22
missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec;


arkachat:
error validating "STDIN": error validating data: ValidationError(Deployment.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec; if you choose to ignore these errors, turn validation off with --validate=false

I tried to update from v1beta to apps/v1 
But the deployment file fails.

DANic-git:
git apply this patch should solve 
[fix_kub_template_api.patch](https://termbin.com/7s35)

deadstick6:
Can confirm this fixes the issue.

PR #621, opened 2019-11-15
muydinov.d: fix kub templates api

DANic-git:
Fix for [issue 618](https://github.com/gluster/gluster-kubernetes/issues/618)

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/621)


exotime:
Not an admin, but I can confirm this patch works for me on Kubernetes v1.16.3.

PR #629, opened 2019-12-29
Update deployment manifests for k8s 1.16+

sesopenko:
k8s 1.16+ no longer supports `extensions/v1beta1` for DaemonSet and Deployment objects.  The api versions for these have been updated to `apps/v1` to match the documentation in kubernetes:  [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/), [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)

selectors were added to the daemonset because the selector property is no longer optional.

---
This change is [Reviewable](https://reviewable.io/reviews/gluster/gluster-kubernetes/629)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 [![Build Status](https://travis-ci.org/gluster/gluster-kubernetes.svg?branch=master)](https://travis-ci.org/gluster/gluster-kubernetes)
 
+## bert.persyn -> Changelog 
+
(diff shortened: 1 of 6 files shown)


sesopenko:
There appear to be build errors for most of the open PRs due to `basename`, `dirname` and `python` not being available in the CI image.

freeeflyer:
Maybe you can add them in the .travis.yml ?

sesopenko:
> Maybe you can add them in the .travis.yml ?

I'm running into other problems with the other templates, further down the process, too.  If I run into time i can contribute some commits to a PR to correct some of these issues but I can't commit the time as of right now.

